PR #150, opened 2025-12-03
Fix Windows Mouse Click Offset

### LINKED ISSUE #159: Clicking buttons in game are offset
COB
Windows
Start new scenario and walk around in COB.
Walk around and wait for random encounters that have the do A or B buttons that appear below the character list. THink the lady vendor with dagger and wizard etc.

Need to aim cursor a little below buttons to actually hit them. The clicker should be the tip of the sword.

danapplegate:
I'm noticing that on Windows specifically, the mouse position appears to be reported by SDL about 16 pixels higher (that is, the `v` is 16 less) than it should be. This causes a couple of problems, one of which is that the cursor icons are sometimes incorrect (hovering over the top left of the camp icon still shows the movement arrow, for instance). The more serious bug is that it makes some interface buttons difficult or impossible to click, such as the bottommost spell in the spell list, or the top character portrait in the trade screen.

~This is a very hacky way to fix this, but there must be a better way to have the Mac and Windows builds receive consistent mouse click coordinates.~

UPDATE: It seems that `SDL_SetWindowSize` actually controls the [client area](https://learn.microsoft.com/en-us/windows/win32/winmsg/about-windows#application-windows) of the Windows window, not the overall window size. Since SDL doesn't provide support for native menu bars, the fact that we have to bypass it and call the Windows API directly to create the menu bar means that SDL doesn't know that the client area has shrunk to accommodate the new menu bar. So, a simple call to `SDL_SetWindowSize` will force it to recompute the total window size necessary to accommodate both the menu bar and the 800x600 client area.

Fixes #159

### DIFF
--- a/src/windows/WinMenuController.cpp
+++ b/src/windows/WinMenuController.cpp
@@ -121,6 +121,14 @@ void WinMenuSync(SDL_Window* sdl_window, std::shared_ptr<WinMenuList> menu_list,
   if (old_menu) {
     DestroyMenu(old_menu);
   }
+
+  // After experimenting with this, it seems that calls to SDL_SetWindowSize actually set the
+  // client area of the window, not the full window size inclusive of the menu bar. Since we have to
+  // bypass SDL to create the menu directly via the Windows API, it seems that SDL doesn't know that
+  // the rendering of the menu bar has shrunk the client area. So, a quick call to SDL_SetWindowSize is
+  // enough to force SDL to realize the menu bar now exists and to expand the window to ensure that the
+  // client area is the full 800x600.
+  SDL_SetWindowSize(sdl_window, 800, 600);
 }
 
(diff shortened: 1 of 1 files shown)


fuzziqersoftware (dismissed):
ChatGPT claims that this is because SDL doesn't know about the menu bar added with SetMenu, which makes sense to me. It also claims that doing the following after SetMenu will fix it, since it will cause SDL to recompute the window's client area:

    RECT r;
    GetClientRect(hwnd, &r);
    SDL_SetWindowSize(sdlWindow, r.right - r.left, r.bottom - r.top);

If this works, it seems like a more correct fix.

danapplegate:
@fuzziqersoftware Unfortunately, it looks like ChatGPT's solution didn't work. But I did notice that, when the menu first appears, the main screen image gets a little squished horizontally. So, perhaps ChatGPT is on the right track, and we need to expand the SDL window to account for the menu height?

jpetrie (approved):

Issue #171, opened 2025-12-07
Bundled Realmz has incorrect rpaths


jpetrie:
I ran into this working on #168, but I've confirmed it happens with the current code as well, although I could be missing something so it would be neat to confirm it independently. 

I believe the bundled version of Realmz generated by CPack has incorrect rpath entries (and consequently, so does the version on the `.dmg` for the current release). This will result in an inability to run the game unless the SDL `.dylib`s happen to be in very specific locations.

To reproduce:

1. Download [the latest release `.dmg`](https://github.com/Realmz-Castle/realmz/releases/download/v8.1.0-beta/Realmz-8.1.0-beta-Darwin.dmg), mount it, and drag Realmz somewhere temporary. I put it in `~` for simplicity.
2. On the command line, run `otool -l` on the executable inside the app (e.g., `otool -l ~/Realmz.app/Contents/MacOS/Realmz`. 
3. Note the values of the `LC_LOAD_DYLIB` entries - they specify SDL should be loaded from `@rpath/libSDL3.0.dylib` and such. 
4. Note the values of the `LC_RPATH` entries - they include references to the user directories of whoever packaged this release. If the SDL libraries aren't found at any of those directories, which just won't exist for most users, the application will not run.
5. Double-click the bundled Realmz app to run it, bypassing Apple's usual security warnings because it isn't signed, and note that you'll get an error that "Realmz cannot be opened because of a problem."
6. Clicking "Report" on this error window brings up the crash report, which indicates `Termination Reason:  Namespace DYLD, Code 1, Library missing` because `Library not loaded: @rpath/libSDL3_ttf.0.dylib`

(If you are @danapplegate, you would need to delete the locally-built versions of the SDL libraries to reproduce this, since there are rpath entries pointing into your user directory where your build environment is).

Repeating the steps on a bundle I built yields the same result, the executable ends ups with rpath entries that point into my user directory where I have my build environment for Realmz. The app will launch fine, up until I delete the SDL `.dylibs` from my build directory, and then it fails.

There is an rpath entry for `@executable_path`, which means that resolving this could be as simple as ensuring the SDL `.dylibs` are installed next to the executable in the bundle. I will also make sure the bundle generation I'm exploring as part of #168 fixes this.

Issue #168, opened 2025-12-06
Launching the non-packaged executable crashes ("current resource file is out of bounds")


jpetrie:
The content of `.vscode/launch.json` suggests that it should be possible to launch the non-packaged build of Realmz by running it from the repository root with the working directory set to the repository root. That is, after building Realmz per the instructions, typing `./build_mac/Realmz` should work. Instead, it crashes.

(This also happens if you use CMake to generate a Xcode build tree, even after manually setting the working directory for the Realmz target because CMake's defaults aren't appropriate.)

The log indicates that this is because:
```
I 60872 2025-12-06 09:49:39 - [ResourceManager] Failed to load resource file :realmz
```

I've been digging around but I'm not entirely sure which file this is supposed to be (my current assumption is "resources/realmz.rsrc" but I'm not sure).

It would be useful to be able to run the built app locally/under a debugger without having to wait for the full package to complete every time.

jpetrie:
What was the motivation for switching away from bundle type targets in the first place?

I haven't used CPack much, but I've gotten pretty far in turning the Realmz target into a bundle (or _back_ into a bundle, apparently :D) and it doesn't seem to impact CPack's ability to generate the `.dmg` for distribution, so it seems like it would be possible to have the bundle target and use CPack to package things up for releases.

veganvelociraptor:
Not sure what I'm doing wrong here. I can get Realmz v7.1.2 running in Windows with no issues, but not any builds from here. I tried copying realmz.rsrc into the Resources directory, and creating a duplicate copy without the rsrc suffix.

This is the output trying with the 8.1.0beta3 package:

```
P:\Downloads\Realmz>realmz
D 3012 2026-06-14 12:07:27 - [CCGrafPort:0000000000325370] Created
D 3012 2026-06-14 12:07:27 - [CCGrafPort:0000000000338760] Created
D 3012 2026-06-14 12:07:28 - [WindowManager] WindowManager::create_sdl_window()
D 3012 2026-06-14 12:07:28 - [CCGrafPort:0000000000338760] Resized to 800x600
Built-in video drivers: windows, offscreen, dummy
Video driver: windows
Built-in render drivers:
  direct3d11
  direct3d12
  direct3d
  opengl
  opengles2
  vulkan
  gpu
  software

Primary display info:
  Name:                 1
  Pixel Format:         16161804
  Width:                1230
  Height:               750
  Pixel Density:        1.000000
  Refresh Rate:         60.000000

[...]

PR #172, opened 2025-12-07
Generate a bundle for local builds.

jpetrie:
Following on the from discussion on issue #168, this PR includes:

 - Generation of an application bundle for local builds of Realmz, ensuring that local tests are against binaries/configurations that mirror released tests as closely as possible.
 - Preservation of the `.dmg` installer generation using CPack (now using the `DragNDrop` package generator instead)
 - Fixup of the rpath entries in both the local and packaged bundles, allowing them to be fully-relocatable (this fixes the issue reported in #171).

Turning the executable target back into a bundle was straightforward, just adding some target properties back and ensuring all the data files had appropriate bundle locations. The `install` target is still valid, but will just install the generated bundle to where CMake is told to install files.

The `package` target was switched to use the `DragNDrop` CPack generator, which simply receives what `install` produces and generates a `.dmg` with an Applications folder symlink like before. This should look identical to end-users.

Fixing up the bundle rpaths was a bit more involved and I feel like there is a more elegant solution here. I left additional comments in context in the commit with some details.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -405,27 +405,6 @@ set_source_files_properties(${CHARACTER_DATA_FILES} PROPERTIES
 
 if(APPLE)
     install(TARGETS Realmz BUNDLE DESTINATION ".")
-
-    # Install data files to <bundle_root>/Resources
-    # install(FILES ${RESOURCE_FILES} DESTINATION ".")
-    # install(FILES ${DATA_RESOURCE_FILES} DESTINATION "Data Files")
-    # install(FILES ${CHARACTER_RESOURCE_FILES} DESTINATION "Character Files")
-    # install(DIRECTORY "${CMAKE_SOURCE_DIR}/base/Realmz/Scenarios" DESTINATION ".")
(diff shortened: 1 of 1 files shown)


jpetrie:
I agree with you on the issue division and relative prioritization, and that it's worth a separate, quicker fix for the RPATH issue.

As far as the developer build process, I believe generating a bundle again is a better approach because it means developers are testing and iterating on something that's as close to what would be shipped as possible.

The alternative I see to generating the bundle would be to have post-build steps copy resources into the output directory and essentially assemble the bundle structure without the top-level `.app` folder of the bundle itself, and I believe that would result in a more complex CMake setup overall then just setting the bundle location property on the relevant files.

The main advantage I see to using CPack to make the bundle is that it seems to directly support code-signing and entitlements, should that ever become desirable. But that could be accomplished with relatively simple post-build steps instead, almost as easily.

jpetrie:
Given that the bundle can still be created by building the appropriate target, I think the better option here to address the underlying issue is to use the file copy solution; closing this.

PR #174, opened 2025-12-09
Fix some warnings

jpetrie:
Nothing groundbreaking, this PR just addresses some of the one-off warnings.

### DIFF
--- a/src/SoundManager.cpp
+++ b/src/SoundManager.cpp
@@ -2,6 +2,7 @@
 
 #include <SDL3/SDL.h>
 #include <algorithm>
+#include <limits>
 #include <memory>
 #include <phosg/Encoding.hh>
 #include <resource_file/ResourceFile.hh>
@@ -28,7 +29,7 @@ public:
     if (this->device_id > 0) {
       return;
     }
-    if (SDL_Init(SDL_INIT_AUDIO) < 0) {
+    if (!SDL_Init(SDL_INIT_AUDIO)) {
       sm_log.warning_f("Could not initialize audio subsystem: {}", SDL_GetError());
       return;
     }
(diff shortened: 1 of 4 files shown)


fuzziqersoftware (approved):

src/SoundManager.cpp
```diff
@@ -77,7 +78,15 @@ class SoundManager {
...
-    if (SDL_PutAudioStreamData(sdlAudioStream, sound->data.data(), sound->data.size()) < 0) {
+    // SDL_PutAudioStreamData takes an int to represent the audio stream size in bytes. It's unlikely that any sound
+    // will exceed that, but just in case, this check makes it obvious. It also guarantees the cast of the size argument
+    // is safe.
+    if (sound->data.size() > std::numeric_limits<int>::max()) {
+      sm_log.warning_f("Audio stream data is too large ({} bytes)", sound->data.size());
+      return;
+    }
+
+    if (!SDL_PutAudioStreamData(sdlAudioStream, sound->data.data(), static_cast<int>(sound->data.size()))) {
```
danapplegate: I guess an alternative is that we could simply clamp the size to max int, and let SDL play as many samples as it can. That would clip the sounds, but at least we would hear something before it's clipped, rather than silently emitting a log message.

A better approach might be to create a buffer queue of max int sized chunks of audio data, start playing the first chunk, then use the `SDL_AudioStreamCallback` parameter of [`SDL_OpenAudioDeviceStream`](https://wiki.libsdl.org/SDL3/SDL_OpenAudioDeviceStream) to queue up additional chunks as playback progresses. This would allow sounds of arbitrary length to be played. I wonder if this might even be necessary once we start re-implementing the music functionality, with long-ish song files and looping playback.

But, for now, I agree with you that this is highly unlikely to matter, and I'm fine with this fix for the warning at the present.

danapplegate (approved):

PR #182, opened 2025-12-10
support menu keyboard shortcuts on macOS; ref #151

fuzziqersoftware:


src/MenuManager.cpp
```diff
@@ -262,4 +288,8 @@ int32_t PopUpMenuSelect(MenuHandle menu, int16_t top, int16_t left, int16_t popU
 int16_t CountMItems(MenuHandle theMenu) {
   auto m = mm.get_menu(theMenu);
   return m->items.size();
-}
\ No newline at end of file
+}
+
+int32_t MenuKey(int16_t ch) {
+  return 0;
```
danapplegate: Did you mean to call the new `find_item_by_key_equivalent` function here?

danapplegate: Nevermind, we talked in person, and this is unnecessary, because Cocoa should intercept the keydown event and call our custom menu callback. Not sure this works on Windows, but we can tackle that separately.

PR #184, opened 2025-12-11
Address pointer conversion warnings for strings.

jpetrie:
This PR proposes changes that address one of the biggest sources of warnings in the code, pointer signedness conversions for strings. This cuts the warning count by half.

The main change here is to switch the various string types in `Types.h` to `char` from `unsigned char`. Because the majority of string usage in the code is actual C strings (which are passed to C standard library functions), this silences almost all of the warnings. For the cases where actual Pascal strings appear (and the length byte is tested), I've made sure to cast the byte to an unsigned value to ensure that any math done with it does not treat it as a signed value. The only place I found in my audit that I left alone was the registration code handling section, since it isn't used.

I opted for this approach because it would be fairly easy to revert locally (just add `unsigned` back to `Types.h` - the various casts can be left alone since they'll become no-ops) if A/B testing is needed, and because the other options that I explored ultimately seemed worse.

For reference, the other options I considered were:

 - Adding macros for the various C string functions that are used to redirect them to functions taking `unsigned char` strings. This isn't good practice, and in fact fails because those functions are _already_ macros in the standard library implementation (to handle redirects to secure versions).
 - Replacing every warning site with a new actual function that addresses the conversion warning. This quickly becomes really intrusive and verbose since most of these call sites are in the original Realmz code, and doesn't fundamentally treat the problem any differently than just redefining the string type. It would also be harder to temporarily undo locally.
 - Clever C++ template tricks to catch C versus Pascal string mismatches at compile time. This of course fails immediately because most of the code is C. Oops.
 - Disabling the warning. This is the least invasive option, but I think it's the worst: for strings, whether or not the characters are signed or not doesn't really matter except for that length byte in the front. But the actual warning applies to pointers of all integer types, and I thought removing that safety net for pointers to integers that are more likely to be used for actual math would be a bad idea.

I tested this by playing through Bywater for an hour or so. Not exhaustive, but I found a few mistakes right of the bat and once I fixed those I didn't see any other issues.

### DIFF
--- a/src/FileManager.cpp
+++ b/src/FileManager.cpp
@@ -121,7 +121,7 @@ OSErr FSpDelete(const FSSpec* spec) {
 }
 
 OSErr FSMakeFSSpec(int16_t vRefNum, int32_t dirID, ConstStr255Param fileName, FSSpecPtr spec) {
-  memcpy(spec->name, fileName, fileName[0] + 1);
+  memcpy(spec->name, fileName, static_cast<uint8_t>(fileName[0]) + 1);
   spec->vRefNum = vRefNum;
   spec->parID = dirID;
   return 0;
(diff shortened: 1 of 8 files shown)


jpetrie:
Closing this PR; we want to focus on gameplay-affecting bug fixes first, and the number of files this change needs to touch would make it unwieldy to merge later versus simply re-creating when we're ready to address the issue. Documented the underlying challenge as issue #271.

PR #186, opened 2025-12-12
Check the right condition when finding secrets.

### LINKED ISSUE #176: checkforsecret.c checks Ogre Hide instead of Discover Secret
checkforsecret.c:26 : partycondition[2] is Ogre Hide; I think this was supposed to be partycondition[3] (Detect Secret)

jpetrie:
This PR fixes #176 by testing the correct party status effect variable ("Discover Secret" instead of "Ogre / Dragon Hide").

### DIFF
--- a/src/realmz_orig/checkforsecret.c
+++ b/src/realmz_orig/checkforsecret.c
@@ -21,7 +21,11 @@ short checkforsecret(short force) /****** 2 = trap/pit one square only, 1 = Secr
 
   chancesecret /= (charnum + 1);
 
-  if (partycondition[2])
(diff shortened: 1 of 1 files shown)


danapplegate (approved):
Nice, thank you! I confirmed that `partycondition[3]` is indeed "Detect Secret" by setting a breakpoint in `resolvespell`, casting Detect Secret, and stepping through it. The [logic here](https://github.com/Realmz-Castle/realmz/blob/081ef9c6e30ce053d7503474e5060e606e8f80d4/src/realmz_orig/resolvespell.c#L112-L113) is responsible for setting it.

PR #188, opened 2025-12-17
Do not double-apply special attacks

### LINKED ISSUE #179: Monster special attacks vs players are applied twice
attack.c:788-858

```
case 6: /**** poison ****/
   if ((!savevs(4, chare)) && (character.condition[9] > -1)) {
     if ((c[chare].condition[9] += abs(specdamage)) < 30)
       c[chare].condition[9] += abs(specdamage);
     sound(630);
     showresults(chare, -28, mon); /****** special attack ******/
   }
```

The use of the += assignment operator instead of == equality testing means that conditions are added, and if the affected character has less than 30 levels of that condition, the comparison evaluates as true and the condition is added a second time. The result is that the actual duration of most conditions caused by melee attacks (fear, paralyze, curse, stupefy, entangle, poison, confuse, disease) are twice what is displayed.

danapplegate:
Fixes #179.

### DIFF
--- a/src/realmz_orig/attack.c
+++ b/src/realmz_orig/attack.c
@@ -791,15 +791,24 @@ trynewweapon:
            *   if ((c[chare].condition[0] += abs(specdamage)) < 30)
            *     c[chare].condition[0] += abs(specdamage);
            *
-           * This appears to have been intended to ensure that the stacked duration of each character
-           * condition could not exceed 30 rounds. However, note that the compound assignment operator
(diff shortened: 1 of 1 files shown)


danapplegate:
Ah, that's a great point. Your proposal would work, and it would continue
the current behavior of allowing the condition duration to exceed 30 while
avoiding applying the damage twice. Reading into the intention of the code
a bit more, it seems like this is an attempt to clamp the maximum duration
at 30, so we could just do that. What do you think?

Issue #185, opened 2025-12-11
Incorrect handling of brawn during character creation


Orthotope:
In newcharacter.c, line 160, age-related modifiers are applied to new characters. For brawn, this involves subtracting the current to-hit and damage bonuses, applying the age modifier, and adding the new bonuses. The problem is that this assumes the character already has those bonuses for their current brawn score, and that doesn't happen until line 234. The result is inconsistent behavior during character creation: bonuses that are higher than expected, do not increase linearly with different brawn values, and in some cases characters with very low brawn will receive bonuses.

The fix for this would be to move lines 234-237 (starts with "strength(characterl.st);") to just above line 160 ("characterl.currentagegroup = 0;").

PR #197, opened 2026-01-13
Apply base to-hit and damage bonuses before aging.

jpetrie:
This change moves the initial to-hit and damage bonus calculation to just prior to the aging modifier application, as suggested by @Orthotope in #185.


danapplegate (approved):
Nice! Thanks for fixing this. Took me a bit to step through the logic and see why it was wrong.

PR #201, opened 2026-01-24
Throttle key repeats to avoid sluggish gameplay.

### LINKED ISSUE #199: Input delay with keyboard movement on macOS
There's a pretty significant input delay one can observe on macOS builds. Start Bywater and navigate to here:
Image

Press and hold the right arrow key, and keep it pressed as your character moves until you get to about here:
Image

Once released, the game will continue to process the queued key down events, and the party will move very far from the spot you released the key (for me, I ended up just past the abandoned well). It results in a fairly laggy feeling play experience.

System key repeat and repeat delay probably exacerbates this (mine is set as below on 26.2):
Image

During this process you can observe the event logging take about three seconds to process all enqueued key down events before it hits the key up.

jpetrie:
This change prevents key down events from being enqueued if they are for a key that has an existing keydown in the queue that occurred sooner than a given threshold. The threshold is set to the amount of time Realmz will wait, via `checkdelay()`, before advancing the game state. Since Realmz uses a busy-wait loop to implement that delay, events aren't processed and can build up in the queue, resulting in a lack of player control as Realmz drains the key repeats.

This would fix #199.

### DIFF
--- a/src/EventManager.cpp
+++ b/src/EventManager.cpp
@@ -9,11 +9,6 @@
 #include "Types.hpp"
 #include "WindowManager.hpp"
 
-// Defined in main.c of the original Realmz code, this value is the base scalar for the game speed. It is used in this
-// file to prevent repeated key down events from clogging the event queue faster than the game will read them; see the
-// code in EventManager::enqueue_sdl_event().
-extern short oldspeed;
-
 static phosg::PrefixedLogger em_log("[EventManager] ", DEFAULT_LOG_LEVEL);
 
 static constexpr uint16_t EVMOD_RIGHT_CONTROL_KEY_DOWN = 0x8000;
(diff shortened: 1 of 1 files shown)


src/EventManager.cpp
```diff
@@ -9,6 +9,11 @@
 #include "Types.hpp"
 #include "WindowManager.hpp"
 
+// Defined in main.c of the original Realmz code, this value is the base scalar for the game speed. It is used in this
+// file to prevent repeated key down events from clogging the event queue faster than the game will read them; see the
+// code in EventManager::enqueue_sdl_event().
+extern short oldspeed;
```
danapplegate: This is causing a linker error for me when building the GraphicsTest executable, because `main.c` isn't included in its source files. Do you have your build environment set up to only build the `realmz.lib` and `Realmz` executable targets?

I'm fine with removing the GraphicsTest entirely. It's not that useful anymore.

src/EventManager.cpp
```diff
@@ -435,10 +440,39 @@ class EventManager {
...
+            auto cursor = event_queue.rbegin();
+            while (cursor != event_queue.rend()) {
+              EventRecord& existing = *cursor;
+              if (existing.what == keyDown && existing.message == message) {
+                // Determine how long it's been since the most recent key down event for this same key.
+                uint32_t duration = TickCount() - existing.when;
+
+                // Note that calculation used for the actual checkdelay() function in Realmz is "2 * oldspeed * scale"
+                // where scale is the parameter to checkdelay(), but Realmz never calls it with a value other than 1.
+                uint32_t threshold = 2 * oldspeed;
```
danapplegate: This works well as-is, but I'm wondering if the duration and threshold calculation based on the game speed are really necessary. Can we get by with simply checking if there's an existing keydown event in the queue already? If the key repeat rate is high enough that it's racing against the game loop, then it will have plenty of chances to insert a new keydown event once the already queued one is processed.

PR #204, opened 2026-01-31
Use userdata directory for preferences file

danapplegate:
This changes our file path mapping code to look for the preferences resource file, and create if missing, in the userdata directory, rather than the default executable directory. This allows the user's preference choices to survive through reinstall cycles. It also avoids problems on systems where the installation directory is not writable by the game, such as on Windows when the game is installed in a directory like "Program Files." Since the preferences file is created if missing, this leads to errors when the game can't write the directory.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,19 +1,30 @@
-*.o
-*.xSYM
-*.SYM
+# Build products / editor files
 *.dSYM
-.DS_Store
-.*.swp
-.*.swo
-
+*.o
+*.SYM
+*.xSYM
 build
-deployment
-*.code-workspace
-
+cmake_install.cmake
 CMakeCache.txt
 CMakeFiles
-Makefile
-cmake_install.cmake
-
+CPackConfig.cmake
+CPackSourceConfig.cmake
+deployment
+GraphicsTest.app
+install_manifest.txt
 librealmz_lib.a
+libSDL3*.dylib
+Makefile
 
+# Editor files
+.*.swo
+.*.swp
+.vscode
+*.code-workspace
+
+# Finder files
+.DS_Store
+
+# Other stuff
+divinity.out
+realmz.out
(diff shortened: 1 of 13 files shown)


src/FileManager.cpp
```diff
@@ -129,11 +138,10 @@ OSErr FSMakeFSSpec(int16_t vRefNum, int32_t dirID, ConstStr255Param fileName, FS
...
 OSErr FindFolder(int16_t vRefNum, OSType folderType, Boolean createFolder, int16_t* foundVRefNum, int32_t* foundDirID) {
   // This is only used by Realmz for getting the Preferences folder (which in
-  // Classic Mac OS is within the System Folder). Here, we just use the local
-  // working directory instead, and we signal this by setting the directory ID
-  // to -1.
+  // Classic Mac OS is within the System Folder). Here, we just use the userdata
+  // directory instead, and we signal this by setting the directory ID to 1.
   *foundVRefNum = 0;
-  *foundDirID = -1;
+  *foundDirID = 1;
```
danapplegate: I had to change this to a positive sentinel value because in `openpref()`, the `dir_ID` parameter that is passed by reference to the `FindFolder` method is a `long`. On modern systems, this is likely to be a 64-bit signed integer. This is a problem, because here, `foundDirID` is a pointer to a signed 32-bit integer. This causes `long dir_ID` to actually contain the value `8589934591` – setting the value to the twos-complement representation of -1 while it was a `int32_t *` had the effect of setting the lower order bits of the long to all 1s. Then, when the value is reinterpreted as a `long`, all 8 bytes are used, but the higher order bytes are all 0s, making it appear to be a positive twos-complement number.

We could also fix this by changing `long dir_ID` in `openpref` to match the `int32_t` of this parameter. But, since this is a sentinel value anyway and can be anything we choose, I figured it's easier to just change to a positive value that doesn't risk reinterpretation.

fuzziqersoftware (approved):

Issue #183, opened 2025-12-11
Discussion: Handling readability-focused code changes


jpetrie:
Quite a lot of the Realmz code is written in a style that uses succinct variable names, which are sometimes difficult to follow immediately. Similarly, there's a _lot_ of scenarios where literals are using to index into arrays representing specific conditions, abilities, statistics et cetera and understanding what the meaning of those indices are is difficult without some outside reference.

Similar to #180, I'm looking for some guidance on this. Are changes aimed at improving code readability, such as expanding abbreviations (`character.lu` to `character.luck`) or defining semantic constants for literals (`character.spec[0]` to `character.spec[SneakAttackChance]`) or simply adding commentary to describe some of the more complex logic something that should be encouraged? Or is the goal to keep the code as true to the original as possible, only making changing when necessary to address bugs or get something working?

jpetrie:
PR #213 adds named constants for various conditions, which establishes a precedent for doing so and is (I think) the right path forward, so I think that resolves this discussion for now.

PR #213, opened 2026-03-28
use enums for character, monster, and party conditions; closes #210, closes #212

### LINKED ISSUE #210: Incorrect displayed values for attack/defense bonus
In attack.c, when calculating the to-hit chance, there are lines like 543:
`att -= abs(character.condition[38]); /*** defense bonus ***/`

The use of abs() ensures proper handling for permanent effects (negative values). However, in showstats.c, the comparable line 79 is:
`temp += characterl.condition[38]; /*** defense bonus ***/`

Without abs(), this adds the raw condition value, which changes the displayed stat in the wrong direction if the condition is permanent.

This doesn't occur very often, as entangled, hindered attacks, hindered defense, and defense bonus are rarely set permanently. Aside from cursed items, it can be most readily seen with Dragon's Claws and the Ring of the Warriors (-5 defense bonus for each).

Note that it's only the displayed value on the character/inventory screen that is incorrect; the actual combat calculations are handled correctly.

Proposed fix: add an abs() call to showstats.c lines 57, 59, 77, and 79.

### LINKED ISSUE #212: Charm Protection party condition not working
In misc.c:1201, there's a 50% bonus to player saves vs. charm (spell type 0) if party condition 7 is active. This is clearly supposed to implement Charm Protection. However, Charm Protection (as set by the spell Thought Lace or the items Robe of Tranquility, Ring of Harmony, or Drums) is party condition 8. Party condition 7 is actually Sentry.

fuzziqersoftware:


jpetrie (approved):
Looks great. I think this also addresses the discussion in issue #183.

PR #214, opened 2026-03-31
fix long vs. int32_t in save file format

fuzziqersoftware:
Also, banish all long ints and replace them with int32_t. See comments in loadsavedgame.c for details of what happened.


src/realmz_orig/loadsavedgame.c
```diff
@@ -331,17 +368,31 @@ short load(void) {
...
+  } else {
+    fread(&testlocation, sizeof testlocation, 1, fp);
+  }
   CvtLongToPc(&testlocation);
 
-  fwrite(&spellcasting, sizeof(Boolean), 1, fp); //*** fantasoft v7.1b  Added as was missing in previous version.
-  fwrite(&spellcharging, sizeof(Boolean), 1, fp); //*** fantasoft v7.1b  Added as was missing in previous version.
-  fwrite(&monstercasting, sizeof(Boolean), 1, fp); //*** fantasoft v7.1b  Added as was missing in previous version.
-  fwrite(&spareboolean, sizeof(Boolean), 1, fp); //*** fantasoft v7.1b  Added as was missing in previous version.
+  fread(&spellcasting, sizeof(Boolean), 1, fp); //*** fantasoft v7.1b  Added as was missing in previous version.
```
jpetrie, line 392: So these values were previously never read correctly (since `fp` is opened in `"rb"` mode above and the `fwrite()` calls would presumably just fail)?

jpetrie (approved):

src/realmz_orig/loadsavedgame.c
fuzziqersoftware: Yeah, it looks like a copy-paste error in the original code. I'm guessing this never actually worked.

src/realmz_orig/loadsavedgame.c
```diff
@@ -284,7 +301,21 @@ short load(void) {
...
+  // NOTE(fuzziqersoftware): The original Realmz Classic format saved 3 64-bit integers here, but only saved the first
+  // 12 bytes of them due to the use of sizeof(int32_t). The byte layout in memory is (G=gold, M=gems, J=jewelry):
+  //   bank = GG GG GG GG GG GG GG GG MM MM MM MM MM MM MM MM JJ JJ JJ JJ JJ JJ JJ JJ
+  // But we only save the first 12 bytes to the file, which is:
+  //   bank = GG GG GG GG GG GG GG GG MM MM MM MM
+  // The correct format in the save file is (truncating off the high 32 bits of each field):
+  //   bank = GG GG GG GG MM MM MM MM JJ JJ JJ JJ
+  // Thankfully, it seems Myriad forgot to byteswap these fields, so we do actually get the low 32 bits of the gems
+  // value in the incorrect case, but the jewelry value is entirely lost. To recover from this, we can simply move the
+  // jewelry value to gems, and clear the jewelry value. (Sorry about that!)
```
danapplegate, line 313: @fuzziqersoftware Sorry, I know I'm super late to this PR. But isn't this a problem of our own making? Originally, the code here (and in `save()`) wrote and read 3 `long` values, which was consistently 12 bytes (3 32-bit ints) in the Classic environment. So, original save files should have 12 bytes here.

However, when we first compiled this code on modern systems and started saving 3 `long`s for the bank data, we were saving 3 64-bit integers. So, some of our modern save files now contained 24 bytes of bank data. Then, we started swapping out `sizeof(long)`s for `sizeof(int32_t)`, which choked on these modern saves.

In short, I think the only save files that would have 24 bytes of bank data and need special treatment here would be the ones that we happened to generate while `save()` was writing 3 `long`s. Does that sound right?

PR #217, opened 2026-04-05
implement menu item enabled state in popups

fuzziqersoftware:
This makes the "Party Members Who Can Use This Item" menu useful again.

### DIFF
--- a/src/MenuManager.cpp
+++ b/src/MenuManager.cpp
@@ -104,18 +104,16 @@ public:
       }
     }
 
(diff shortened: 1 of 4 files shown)


jpetrie (approved):

danapplegate:
I tried testing this on my latest build, but I'm not sure it's working. I went to a shop, opened up the details page for a longsword, clicked the "Party Members Who Can Use This Item" button, but it shows my magic users like my Evoker and Priest can use it. Same on both Mac and Windows. Maybe I'm missing something?

PR #225, opened 2026-04-22
Show bundled characters in party select, fix resulting Data CD crash

### LINKED ISSUE #161: Begin new adventure screen has issues (windows)
Windows. 
Click from menu "Begin new adventure".

Click outsdie the list of 6 default characters on the grey area they appear in to add to party and you get the popup "Please make sure the file is in the folder". Sometimes this also makes one of the characters dissappear. Restarting realmz does not necesarily mean they reappear again.

I'm down to 3 character files doing this. Files still exist in the folder, just not in the list.

chromancer:
#208 does character discovery via directory enumeration (but left Data CD). Bundled characters from original implementation no longer show as a result. If mac_list_directory is made symmetric with mac_fopen's userdata+bundle fallback, freadx<struct character> chokes on Data CD.

This PR:

- mac_list_directory unions userdata and bundle dirs, deduping by name, user data wins.

- Delete Data CD.

This may partially fix #161, but can't test on Windows


danapplegate (approved):
Thanks @chromancer!

PR #226, opened 2026-04-22
stop character hiding on empty slot click

chromancer:
issue: Clicking empty slot while any character is selected triggers double popups and hides the selected character from the list (until quit/relaunch in current build). This is original behavior.

Empty slot > empty text > fopen fails > calls minus(filepick_text, 1), which hides wrong character.

Fix: guard minus() call on clicked slot having had a name. 

note warn() stomps myString, so stashed empty-check in a local.

I left the annoying popups for nostalgia purposes. If you want to lose that behavior also it can be done but it isn't original.

This is other half of fix for #161 I think. Can't reproduce on Windows. This plays clean on my build.

### DIFF
--- a/src/realmz_orig/partyselect.c
+++ b/src/realmz_orig/partyselect.c
@@ -395,15 +395,24 @@ shortupdate:
       strncat(filename, myString, 30);
 
       if ((fp = MyrFopen(filename, "rb")) == NULL) {
+        /* *** CHANGED FROM ORIGINAL IMPLEMENTATION ***
+         * NOTE(chromancer): Bug in original code hides selected character
+         * when fopen fails, which vanishes the selected on an empty-slot
+         * click. Fix: guard on the clicked slot having had a name. warn()
+         * overwrites global myString, so capture check before warn().
+         */
+        short clicked_empty = !strlen(myString);
         SetPortDialogPort(party);
         ForeColor(blackColor);
         BackColor(whiteColor);
         warn(38);
         warn(143);
-        GetDialogItem(party, filepick, &itemType, &itemHandle, &itemRect);
-        GetDialogItemText(itemHandle, myString);
-        PtoCstr(myString);
-        minus((Ptr)myString, 1);
+        if (!clicked_empty) {
+          GetDialogItem(party, filepick, &itemType, &itemHandle, &itemRect);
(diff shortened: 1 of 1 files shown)


src/realmz_orig/partyselect.c
```diff
@@ -395,15 +395,19 @@ void partyselect(short mode) {
       strncat(filename, myString, 30);
 
       if ((fp = MyrFopen(filename, "rb")) == NULL) {
+        /* Changed from original. Capture before warn(), prevent clicking empty slot making selected character vanish. */
```
danapplegate: Would you mind updating the comment here to follow the format we've used in other places, e.g. [here](https://github.com/Realmz-Castle/realmz/blob/fbfca762c2c7aa54cc86fa8645c08913824244c6/src/realmz_orig/name.c#L92-L94)? Just to make it easier to search for all the places where we've modified the original code.

danapplegate (commented):
@chromancer Awesome, thank you! I've confirmed that this fixes the behavior of removing character names from the list.

Our loose policy has been to only fix behavior that is clearly a bug in the original code, taking into account intention and common programming pitfalls, and I think this behavior definitely falls under that category. Errors in scenario action point handling, or where the original intention of the code is unclear, are more likely to be considered original behavior and left as-is.

I think the annoying double warn behavior if you accidentally click an empty slot could also be considered an original bug, but we can leave that discussion for a separate PR.

src/realmz_orig/partyselect.c
chromancer: This look cool to you?

Personally I think the double warn is kind of cute so I intentionally left it in, but if it bothers somebody, no skin off my back.

This project is genius btw. Glad somebody is doing it and glad I can help in some small way. This is unrelated but there is a discord invite link somewhere in the project that doesn't work. I'd be glad to hop on there if it's still alive.

danapplegate (approved):

danapplegate (approved):

src/realmz_orig/partyselect.c
danapplegate: Here you go! Still quite active, and we have a #realmz-classic channel dedicated to this project specifically. https://discord.gg/5UXwTfs7

Issue #156, opened 2025-12-05
it'd be nice to support icons in menu items (low priority)


fuzziqersoftware:

chromancer:
Closed as completed by #264 and #230.

PR #230, opened 2026-04-24
icons restored to Mac menus and dropdowns

chromancer:
This PR contains Mac-side support for menu and dropdown items with cicns (#156)

- Store absolute cicn resource ids on Menu::Item, converting at parse
- Add DecodeCIconImage to get RGBA
- NSImage with 2x so they're not tiny, attach to Cocoa controller
- SetItemIcon now implemented

Despite touched files I didn't do anything Windows-side, just rename the field to be more inline with how it comes out now.

### DIFF
--- a/src/macos/MenuController.mm
+++ b/src/macos/MenuController.mm
@@ -4,10 +4,62 @@
 #include <cstddef>
 #include <cstdint>
 #include <objc/NSObject.h>
-#include <resource_file/ResourceFile.hh>
+#include <phosg/Image.hh>
 
 NSMenu* MCCreateMenu(const MenuList& menuList);
 NSMenu* MCCreateSubMenu(NSString* title, const Menu& menuRes, const std::list<std::shared_ptr<Menu>> submenus);
+
+// We have to forward declare because of PixMap collision w/QuickDraw
+phosg::ImageRGBA8888N DecodeCIconImage(int16_t iconID);
+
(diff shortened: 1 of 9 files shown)


src/macos/MenuController.mm
```diff
@@ -4,11 +4,63 @@
...
+    if (rep == nil) {
+      return nil;
+    }
+    const uint32_t* src = decoded.get_data();
+    uint32_t* dst = reinterpret_cast<uint32_t*>([rep bitmapData]);
+    for (NSInteger y = 0; y < dstH; y++) {
+      const uint32_t* srcRow = src + (y / kMenuIconUpscale) * srcW;
+      uint32_t* dstRow = dst + y * dstW;
+      for (NSInteger x = 0; x < dstW; x++) {
+        dstRow[x] = __builtin_bswap32(srcRow[x / kMenuIconUpscale]);
```
fuzziqersoftware: Is this specific to little-endian architectures, or does addRepresentation always expect ABGR8888? If the former, we probably should condition this on the system's byte order. (phosg::IS_BIG_ENDIAN could be used for expediency here)

fuzziqersoftware (approved):

src/macos/MenuController.mm
chromancer: I think this is what you mean? Updated.

chromancer:
Note I added a commit to drop this sync, see #235 for details

danapplegate:
Just tested on Mac, seems to work well. Awesome, thanks!

PR #231, opened 2026-04-27
enable PC magic resistance

### LINKED ISSUE #228: magic resistance might be bad to have
resist.c line 70, in the party-only branch
```
else if (Rand(100) <= c[who].magres + (powerlevel * spellinfo.resistadjust))
```
magic resistance (magres) is a positive integer. It's added to the caster's power level times the magic resistance adjustment of the spell (negative or zero) to create a percentage change the spell is resisted. So, we roll a d100, use that %, and if we hit it <=…

line 71
```
      return (FALSE);
```
uh oh!

Seems the net gameplay effect is that magic resistance is a percentage chance that you are *hit* by the spell (as might be anyway), and, as a bonus, since it returns false immediately rather than reply = TRUE, it overrides any protection from magic spells you have on you at the time.

I think this is a simple inversion, this should have been `return (TRUE);`

I raise it here as an issue because there are 2 downstream effects on gameplay consonant with #180 :

- game gets easier. Magic resistance has been (for as long as this bug has existed?) either doing nothing—not changing the result of the spell—or directly overriding any protection you had. Magic resistance will now work as intended, which means PCs do get a small chance to resist the spell, and magres doesn't just stomp on your protection from spell level stuff.
- PCs with race/caste bonuses to magres are now considerably better than they were.

Changing this does not affect DRVs caster/magic/fire/cold etc. This purely affects magic resist and the effectiveness of protection spells. I should also note that this has potentially been here forever since it's really hard to reason from *absence*; the only visible sign that this isn't working is the absence of your characters showing the "magic resisted" thing every once in a while.

chromancer:
This PR would close #228 by enabling PC magic resistance.

### DIFF
--- a/src/realmz_orig/resist.c
+++ b/src/realmz_orig/resist.c
@@ -68,7 +68,12 @@ short resist(short who) {
       if (Rand(100) <= (c[who].dodge - (spellinfo.tohitbonus)))
         return (TRUE); /**** Missile dodge ****/
     } else if (Rand(100) <= c[who].magres + (powerlevel * spellinfo.resistadjust))
-      return (FALSE);
+    /* *** CHANGED FROM ORIGINAL IMPLEMENTATION ***
+    * NOTE(chromancer): Original code inherited had return (FALSE), meaning magres never did anything,
+    * except overriding the 'magic screen' (Protection from Nth Level) conditions, causing PCs to get
+    * hit when they would otherwise be immune. Note monster magres still worked.
+    */
+      return (TRUE);
   } else {
     for (ttt = castlevel; ttt < 5; ttt++)
       if (monster[who - 10].condition[ttt + 16])


afkelsall:
I was thinking about this change and the comment about how it may change the game too much/make it easier. With the discussion in discord about adding port settings  (new menu or settings.json file), I thought something like this could go in an experimental mode where it's disabled by default but available for people to try and provide feedback on. 

This way the community of players can test these changes and decide if it breaks the nostalgic experience for them. 

Either way it can remain as an option for players to choose and the result of the conversation is about whether to leave it off by default or enable it by default. 

This way we can keep merging changes like this without worrying about the impact/deviating from the classic experience so much.

chromancer:
I should note that my thinking below has evolved a bit from my original issue open where I concluded that this bug was in Realmz forever. I don't actually know that.

So here's the question about that. I _don't know_ what the classic experience was here. I _seem_ to recall that at some point, player magic resistance worked! That's a dim memory from a version I spent the most time playing (7, I think). But under the current "Realmz Classic" port experience, it doesn't work at all; it's actively bad to have.

So the question is, can anyone confirm (and we don't have the source for prior versions) if magres ever worked? I'm going to assume that at some point it did in fact work, unless we take a poll and determine that nobody has ever seen a PC resist magic.

If it did in fact work, the current port behavior is a regression, which is not unexpected given the provenance of the source it's based on.

jpetrie:
The check `if(Rand(100) <= c[who].magres + (powerlevel * spellinfo.resistadjust))` caught my attention here. This means a character's effective chance to resist a spell is their baseline _plus_ the spell's resist adjustment (times the casting power). But the Divinity manual says that a spell's "+/- resist adjustment" makes spells "harder or easier" to resist, implying that that this check should `magres - (powerlevel * spelling.resistadjust)` as well.

EDIT: Per [Discord](https://discord.com/channels/945287853762564156/1441822880899010631/1513706977275154583), spells that are harder to resist have negative adjustment values, making this check work as expected.

jpetrie (approved):

PR #232, opened 2026-04-28
fix: skip incorrectly-sized character files during party selection

### LINKED ISSUE #227: update_character_files_list crashes on stray files
party select tries everything in Character Files. Any stray file crashes party select.

I dodge this in #225 deleting Data CD which seems in line with current intention, but any file there will also crash party select.

Skip files that aren't sizeof(struct character)?

I have no opinion but thought worth raising

jpetrie:
This fixes #227 by catching exceptions from `phosg::freadx` (thrown when the number of bytes it reads does match the expected size of the character record) and ignoring those character files.

### DIFF
--- a/src/realmz_orig/partyselect-helpers.cpp
+++ b/src/realmz_orig/partyselect-helpers.cpp
@@ -5,6 +5,7 @@
 #include <filesystem>
 #include <format>
 #include <phosg/Filesystem.hh>
+#include <phosg/Strings.hh>
 #include <string>
 #include <unordered_set>
 #include <vector>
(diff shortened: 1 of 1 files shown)


chromancer:
if you like I have a little to add to this now or later (https://github.com/chromancer/realmz/tree/addl-char-file-handling)

there are two more active crashes in party select from my original issue, one I mentioned and one I didn't see until recently.

1) if an oversize non-character file exists and is added to party. freadx reads first 872b in this case, assumes it's fine.  the trailing garbage is parsed as a character and crashes downstream.
2) if a subdirectory exists. mac_fopen returns null for directories (correct behavior for emulating classic). but passing a null FILE* to fread is immediate segfault.

the commit adds two add'l guards for the above.

my principle is that while these result from things that *should not* happen (there's no intended reason to have directories in Character Files, nor is there any good reason to add a non-character file to your party) we should guard an inexperienced user against crashes by eliding junk files completely and skipping directories in enumeration.

fuzziqersoftware (approved):
Sorry I missed this one for so long!

PR #233, opened 2026-04-30
doc: add contribution guidance

### LINKED ISSUE #180: Discussion: Handling gameplay-affecting bug fixes
Discussion in the Discord that led to the filing of #176 got me wondering about other bugs that impact gameplay balance. For example, there was ongoing discussion about a likely bug in `attack.c:834` and adjacent areas that doubles up the application of damage.

Fixing certain kinds of gameplay bugs could impact balance by making certain encounters harder or easier, certain items more or less powerful, and so on. Since the goal of this project, as I understand it, is to preserve the classic experience and make it available on modern machines, I'd be interested in getting perspective and guidance from the maintainers on addressing gameplay bugs. Are there certain kinds of bugs that should be left unfixed to preserve the experience, for example?

For my part, what I'd consider a _long-term_ ideal is a way to opt-in to classes of balance/bug-fixes via some kind of configuration (similar to how the Daggerfall Unity project has a "small dungeons" opt-in). But  a configuration system like that requires a fair bit of work and isn't terribly high priority. So in the short term, I'd propose amending the `CHANGED FROM ORIGINAL IMPLEMENTATION` tag that already seems to be in use on edits to the code under `realmz_orig` so that gameplay-affecting bug fixes are marked as `CHANGED FROM ORIGINAL IMPLEMENTATION - GAMEPLAY` or similar. That makes them easy to identify in the event that there's any desire in the future to do something with them.

jpetrie:
A minor update to the README to point potential contributors at discussion issues so those issues themselves don't need to stay open indefinitely.

Closes #180.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,17 @@ On Windows, you can either use the installer wizard for automatic installation, 
 - Attach the crash report and archive of your userdata directory
 - List the steps necessary to reproduce the bug
 
-# Building on Mac
+# Contributing
+
+Pull requests are welcome. Please keep the following in mind as you author changes:
+
+ - When modifying code under `src/realmz_org`, please include comments detailing changes from the original
+   implementation ([example](https://github.com/Realmz-Castle/realmz/blob/fc143ecb7d54b1f7be3ff7e714fea450297b8bb9/src/realmz_orig/warn.c#L61)).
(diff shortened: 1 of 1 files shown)


danapplegate (approved):
Thanks Josh!

Issue #149, opened 2025-11-30
Check marks on menus don't update after selecting a scenario


fuzziqersoftware:
Selecting a scenario does work, though.

danapplegate:
Ah yeah, this should be fixed in the latest build and on main.

PR #235, opened 2026-05-01
add item marks & remove sync-on-change hangs

### LINKED ISSUE #200: macOS: Infinite loading spinner on Tutorial
Hello! It appears the tutorial isn't working on macOS - I get an infinite "beach ball" loading spinner and the application becomes unresponsive.

**Steps to reproduce:**

- Select Adventure > Tutorial
- Select Adventure > Begin New Adventure
- Add one party member
- Infinite loading spinner encountered on the "Loading and sorting bestiary" phase

macOS version: Sequoia 15.7.3
Realmz version: 8.1.0-beta2

chromancer:
combined add feature & hang fix here because they're dependent on one another.

issue: During work on adding menu item marks back (because once I put back the icons on the dropdowns it bothered me) I found cause of hang on tutorial, and fixing this also fixes sync on marks.

cause: anything that changed the main menu bar (checkitem, setitemmark, setmenuitemtext, etc.) called mm.sync which rebuilds the whole menu bar from scratch. The tutorial as a special case doesn't have the prebuilt menu data > runs a sort on the bestiary menu items to build it > every change rebuilds the menu bar very, very, very many times > indefinite hang. This was kind of a deadfall trap for future scenarios.

fix:

- strip syncs from those
- use existing DrawMenuBar sync instead.
- Requires us to add one additional DrawMenuBar in the original source, or item checks lag user actions by one (as in #149); I considered some other options but honestly this was simplest.
- finally, restored Cocoa rendering of dropped menu marks. 

notes: Diamonds are implemented with the dash (NSControlStateValueMixed).
I see no reason this wouldn't fix the tutorial hang on Windows but I've have done no testing.
No Windows marks yet.

Should fix #200. Partially addresses #202.

### DIFF
--- a/src/macos/MenuController.mm
+++ b/src/macos/MenuController.mm
@@ -88,8 +88,12 @@ NSMenu* MCCreateSubMenu(NSString* title, const Menu& menuRes, const std::list<st
         id menuIdentifier = [[MCMenuItemIdentifier alloc] initWithRawIds:menu.menu_id itemId:itemId];
         [subMenuItem setRepresentedObject:menuIdentifier];
         subMenuItem.enabled = subMenuItemRes.enabled;
-        if (subMenuItemRes.checked) {
+        bool mark_is_submenu_link = (subMenuItemRes.key_equivalent == 0x1B);
+        char mark = mark_is_submenu_link ? 0 : subMenuItemRes.mark_character;
+        if (subMenuItemRes.checked || mark == 19) {
           subMenuItem.state = NSControlStateValueOn;
+        } else if (mark != 0) {
+          subMenuItem.state = NSControlStateValueMixed;
         }
 
         // Submenu ids are specified by the itemMark field if the itemCmd field has
--- a/src/MenuManager-C-Interface.h
+++ b/src/MenuManager-C-Interface.h
@@ -27,6 +27,8 @@ int32_t MenuSelect(Point startPt);
 void DisableItem(MenuHandle theMenu, uint16_t item);
 void EnableItem(MenuHandle theMenu, uint16_t item);
 void CheckItem(MenuHandle theMenu, uint16_t item, Boolean checked);
+void SetItemMark(MenuHandle theMenu, int16_t item, int16_t markChar);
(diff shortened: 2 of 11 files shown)


src/macos/MenuController.mm
```diff
@@ -88,8 +88,12 @@ - (NSMenu*)MCCreateSubMenu:(NSString*)title parentMenu:(const Menu&)menu submenu
         id menuIdentifier = [[MCMenuItemIdentifier alloc] initWithRawIds:menu.menu_id itemId:itemId];
         [subMenuItem setRepresentedObject:menuIdentifier];
         subMenuItem.enabled = subMenuItemRes.enabled;
-        if (subMenuItemRes.checked) {
+        bool mark_is_submenu_link = (subMenuItemRes.key_equivalent == 0x1B);
+        char mark = mark_is_submenu_link ? 0 : subMenuItemRes.mark_character;
+        if (subMenuItemRes.checked || mark == 19) {
```
jpetrie, line 93: What does 19 signify here (like is it a special kind of checkmark, why do we need to test for it versus just `.checked`) and can we use a named constant for it?

jpetrie: Oh, I see, it's the value that Realmz code uses when it calls `SetItemMark`. Would be worth noting that probably, but it's not a showstopper.

jpetrie (changes requested):
I'm not sure the one extra `DrawMenuBar()` call at the end of the menu choice handler is sufficient to counteract all the `mm.sync()` removals (though I do agree that removing those syncs should avoid the update cascade bug).

For example: Start Bywater and go to the blacksmith shop at (10, 6), choose to avenge his son and click through the rest of the dialog and the "you got a map" notice. If you then immediately open the Maps/Notes menu, the map isn't there. It doesn't show up until you choose some _other_ menu item (like a Bestiary entry), which triggers the new `DrawMenuBar()` sync.

chromancer:
Per our discord discussion, went through and found the remaining 5 sites. Most of these just needed `DrawMenuBar`, one needed a single line moved.

jpetrie (approved):

PR #236, opened 2026-06-08
Fix non-functional volume/speed menus on Windows

### LINKED ISSUE #206: Volume and Game Speed Submenus on Windows

afkelsall:
The Volume and Speed pickers did nothing on Windows because their Classic Mac hierarchical submenus were never built. MCSync copied only the top level menus and dropped MenuList::submenus, and the Windows menu builder laid each menu out as a flat list, ignoring the 0x1B/mark_character marker that links a parent item to its child menu, so there was no submenu to open.

This carries the submenus across in MCSync and adds a recursive BuildMenu that attaches each referenced submenu with MIIM_SUBMENU and suppresses the spurious Ctrl+0x1b accelerator the marker would otherwise create. It also renders item marks: now that SetItemMark stores the current selection in mark_character, BuildMenu draws a check on the marked Volume, Speed or Music level, collapsing the diamond and dash glyphs to a single check since native Windows menus have no mixed state.

The change is confined to src/windows, which compiles only under WIN32. macOS already builds these submenus and renders their marks in MenuController.mm and is left untouched, so both platforms keep working.

Closes #206. Related to #202.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## Unreleased
+
+- Show the selected level in the Windows Volume, Speed and Music menus by rendering item marks.
+  - CHANGED FROM ORIGINAL IMPLEMENTATION: native Windows menus have no "mixed" state, so every
+    nonzero mark glyph (19 diamond, -41 dash) collapses to a single checkmark.
+
 ## [v8.1.0-beta2](https://github.com/Realmz-Castle/realmz/releases/tag/v8.1.0-beta2)
 
 - Add CMake presets for macOS by @jpetrie in #167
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## Unreleased
+
+- Show the selected level in the Windows Volume, Speed and Music menus by rendering item marks.
+  - CHANGED FROM ORIGINAL IMPLEMENTATION: native Windows menus have no "mixed" state, so every
+    nonzero mark glyph (19 diamond, -41 dash) collapses to a single checkmark.
```
danapplegate: These can be removed, we typically only update the CHANGELOG upon release.

danapplegate (approved):
Excellent! I had a branch in progress that was working on this very thing, so I'm glad you were able to push this over the finish line.

Issue #158, opened 2025-12-05
size of display


mapplegate27205:
The current size of the display makes it hard to see some things.  Can we scale it to about twice it's current size?

danapplegate:
Thanks! Yeah, we've been thinking about adding a hi-res mode to make up for the fact that modern pixel sizes make it very difficult to play the game at its original 800x600 resolution. Will probably have to save this work for the next minor version release, though.

chromancer:
This issue has been addressed by the merge of #256!

jpetrie:
Closed with #256.

PR #237, opened 2026-06-08
Add configurable window scaling

afkelsall:
See this github issue: #158

Realmz renders at a fixed 800x600 normally and is pretty small on modern displays. This adds an optional settings.json next to the executable with a scale multiplier and a filter. It uses a settings file rather than new in-game menu options to keep the classic menus unchanged. The default ships with scale 1.0, so nothing changes unless a player edits it.

The game still renders into its 800x600 buffer; only the final present is scaled via SDL logical presentation, and mouse input is converted back so clicks stay accurate. The filter controls the upscale (auto, nearest, pixelart, linear); auto keeps the bitmap fonts crisp. It stays macOS and Windows compatible: the scaling is cross-platform SDL, and the only platform-specific change restores the Windows client area after the native menu bar is added.

Three options to compare below. IMO pixel art at 1.5x is my personal preference here though linear does make the quest text look better but the "Items" and "Conditions" bitmap font buttons worse. Open each image in a new tab in order and tab between them:
```
    "scale": 1.5,
    "filter": "nearest"
```

image

```
    "scale": 1.5,
    "filter": "pixelart"
```

image

```
    "scale": 1.5,
    "filter": "linear"
```

image

### DIFF
--- a/settings.json
+++ b/settings.json
@@ -1,0 +1,15 @@
+// Realmz settings. Edit the values below and restart the game.
+{
+  "scaling": {
+    // "scale" is a size multiplier applied to the original 800x600 window.
+    // 1.0 keeps the original size, 2.0 makes the window 1600x1200, and so on.
+    // Any value from 1.0 to 4.0 is allowed, including fractional ones like 1.5.
+    "scale": 1.0,
+
+    // "filter" controls how the image is scaled up to fill the window:
+    //   "auto"    - crisp at whole-number scales, smooth at fractional ones (default)
+    //   "nearest" - always crisp, with blocky pixels
+    //   "linear"  - always smooth, with softer edges
+    "filter": "auto"
+  }
+}
--- a/settings.json
+++ b/settings.json
@@ -7,9 +7,10 @@
     "scale": 1.0,
 
     // "filter" controls how the image is scaled up to fill the window:
-    //   "auto"    - crisp at whole-number scales, smooth at fractional ones (default)
-    //   "nearest" - always crisp, with blocky pixels
-    //   "linear"  - always smooth, with softer edges
+    //   "auto"     - nearest at whole-number scales, pixelart at fractional ones (default)
+    //   "nearest"  - sharp pixels, but uneven widths at fractional scales
+    //   "pixelart" - sharp text and edges that stay even at fractional scales
+    //   "linear"   - fully smoothed; softens the bitmap UI text
     "filter": "auto"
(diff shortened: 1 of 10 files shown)


src/Settings.hpp
```diff
@@ -0,0 +1,19 @@
+#pragma once
+
+#include <SDL3/SDL_surface.h>
+
+// Settings loaded from settings.json in the directory containing the
+// executable. The game always renders into a fixed 800x600 logical buffer;
+// these control how that buffer is presented to the screen so the window can
+// be larger than 800x600 on modern displays.
+struct RealmzSettings {
```
fuzziqersoftware, line 9: I feel like we should call this something different - maybe RealmzExtraSettings - to differentiate it from the Preferences window settings (even though those are managed very differently).

src/Settings.cpp
```diff
@@ -0,0 +1,77 @@
...
+      settings.scale_mode = is_integer_scale ? SDL_SCALEMODE_NEAREST : SDL_SCALEMODE_PIXELART;
+    }
+  } catch (const std::exception& e) {
+    settings_log.warning_f("Could not parse {} ({}); using default window settings", path, e.what());
+    return RealmzSettings{};
+  }
+
+  const char* mode_name = (settings.scale_mode == SDL_SCALEMODE_NEAREST) ? "nearest"
+      : (settings.scale_mode == SDL_SCALEMODE_LINEAR)                    ? "linear"
+                                                                        : "pixelart";
```
fuzziqersoftware, line 74: I think this would be better formatted as something like
```cpp
const char* mode_name = (settings.scale_mode == SDL_SCALEMODE_NEAREST)
    ? "nearest"
    : (settings.scale_mode == SDL_SCALEMODE_LINEAR)
    ? "linear"
    : "pixelart";
```

settings.json
```diff
@@ -0,0 +1,16 @@
+// Realmz settings. Edit the values below and restart the game.
```
fuzziqersoftware, line 1: This wording is a bit awkward - maybe "Restart the game after editing the values below to make them take effect"?

fuzziqersoftware (approved):
Eventually we should add a way to configure this in-game, but I don't think that needs to happen now (or even soon - I might do it myself sometime later).

afkelsall:
With the changes @chromancer is doing around the scaling menu I'm inclined to close this PR and let all this go through with his work.

PR #240, opened 2026-06-09
SDL #13134 fix

chromancer:
issue: after using a Mac native popup menu (class, race, condition, etc.), next mouse click does nothing

cause: when a popup opens on mouse-down, macOS routes mouse-up to menu, not app. SDL mishandles, assumes button still held, eats next click. See SDL bug [#13134](https://github.com/libsdl-org/SDL/issues/13134)

fix: on Mac only, flag before opening popup and fake a mousedown if necessary

effects and thoughts: kludge required by our vendored SDL (dev snapshot). SDL already fixed in 3.4.0 ([commit 92eaa34](https://github.com/libsdl-org/SDL/commit/92eaa34277f0711696d366742735fe2a2c974123)).

Alternatively, we could just bump SDL, which contrary to my expectations doesn't seem like a big deal here. I'm gonna do it in my build and see what happens.

testing: Fixes for MacOS 26.5.1, all other platforms opted out.

### DIFF
--- a/src/EventManager.cpp
+++ b/src/EventManager.cpp
@@ -343,6 +343,10 @@ public:
     this->modifier_flags |= EVMOD_MOUSE_BUTTON_UP;
   }
 
+  void expect_orphan_mouseup() {
+    this->expecting_orphan_mouseup = true;
+  }
+
   inline const Point& get_mouse_loc() const {
     return this->mouse_loc;
   }
@@ -372,6 +376,7 @@ protected:
   Point mouse_loc = {0, 0};
   uint16_t modifier_flags = EVMOD_MOUSE_BUTTON_UP | EVMOD_WINDOW_ACTIVATED;
   std::deque<EventRecord> event_queue;
+  bool expecting_orphan_mouseup = false;
 
   void set_modifier_value(uint16_t what, bool enabled) {
     if (enabled) {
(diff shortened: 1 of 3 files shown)


jpetrie:
Does bumping to a version of SDL without this bug require us to bump our minimum macOS deployment target at all?

Generally speaking, I think it would be preferable to take a new dependency version as long as it doesn't introduce any other significant changes, versus putting in a workaround that we'll to remember to undo at some point when we _do_ take a new version.

chromancer:
No—looked into that already, and there's no reason obvious from docs that we couldn't. That said, I haven't thoroughly tested 3.4, and felt loath to suggest it as the solution without doing a bunch of homework on what SDL has been up to that I didn't want to do.

What's your thought?

fuzziqersoftware:
If it's not too hard to upgrade SDL, I think we should try that out. I generally trust SDL to not do dumb / bad things between versions, so as long as we don't drop support for any macOS versions as Josh said, it should should be fine to just upgrade, check that basic graphics/sound/input work, and assume everything will be fine.

chromancer:
Sounds good. Given the apparent consensus here, I'm investigating bumping to 3.4.10, which will contain a fix for this, and a fix for another position-drift issue that came up in my work on the port menu.

PR #241, opened 2026-06-09
Load saved games in windows  from the original retail Realmz on MacOS 9

afkelsall:
Saved games created by the original retail Realmz fail to load: the loader rejects their "Data I1" file with "error 7" because it does not match either known size.

These saves store the single-short `cancamp` field as an array of 10 shorts, an original buffer overflow that is part of the on-disk format, so the file is 18 bytes larger (`0x398B` instead of `0x3979`). Separately, `bank` and `templecost` are the only multi-byte fields the original code never byteswaps; the data is big-endian, so on a little-endian build they load as huge garbage values (a few-thousand-gold bank reads as billions).

Detect the larger format by size and read `cancamp` as 10 shorts into a local buffer, keeping only the first value to avoid the overflow. Byteswap `bank` and `templecost` when reading that format, and write both the 10-short `cancamp` and big-endian `bank`/`templecost` when saving so the layout matches the original and round-trips. Saves written by earlier versions of this port still load. Every other multi-byte field already uses the same big-endian helpers, so macOS and Windows builds behave identically.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## Unreleased
 
-- Load saved games created by the original retail Realmz. These store the cancamp field as 10 shorts, making them 18 bytes larger than the saves this port wrote, so they were previously rejected. The port now reads that format and writes it as well.
+- Load saved games created by the original retail Realmz. These store the cancamp field as 10 shorts, making them 18 bytes larger than the saves this port wrote, so they were previously rejected. The port now reads that format and writes it as well. The bank balance and temple cost in those saves are also now byteswapped correctly, so they no longer load as huge garbage values on a little-endian machine.
 
 ## [v8.1.0-beta2](https://github.com/Realmz-Castle/realmz/releases/tag/v8.1.0-beta2)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,3 @@
-## Unreleased
-
-- Load saved games created by the original retail Realmz. These store the cancamp field as 10 shorts, making them 18 bytes larger than the saves this port wrote, so they were previously rejected. The port now reads that format and writes it as well. The bank balance and temple cost in those saves are also now byteswapped correctly, so they no longer load as huge garbage values on a little-endian machine.
-
 ## [v8.1.0-beta2](https://github.com/Realmz-Castle/realmz/releases/tag/v8.1.0-beta2)
 
 - Add CMake presets for macOS by @jpetrie in #167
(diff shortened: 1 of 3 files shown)


fuzziqersoftware (approved):

fuzziqersoftware (approved):

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## Unreleased
```
jpetrie: We seem to just generate this file from the changelist that GitHub produces on a new release, so I don't think this change is needed.

src/realmz_orig/save-direction-order.c
```diff
@@ -439,13 +446,20 @@ void save(short mode) {
...
+   * out of a local buffer. This keeps our saves the same size as the originals and avoids
+   * the overflow of reading past the single cancamp variable.
    */
   // CvtTabShortToPc(&cancamp, 10); // Myriad ????
   // fwrite(&cancamp, sizeof cancamp, 10, fp);
   // CvtTabShortToPc(&cancamp, 10); // Myriad ????
-  CvtTabShortToPc(&cancamp, 1); // Myriad ????
-  fwrite(&cancamp, sizeof cancamp, 1, fp);
-  CvtTabShortToPc(&cancamp, 1); // Myriad ????
+  {
```
jpetrie, line 457: This should have a "changed" comment marker as well. Everything else in this file seems close enough to an existing change marker.

src/realmz_orig/loadsavedgame.c
```diff
@@ -118,11 +118,22 @@ short load(void) {
   fseek(fp, 0, SEEK_END);
   size_t file_size = ftell(fp);
   fseek(fp, 0, SEEK_SET);
+  // The original retail builds wrote `cancamp` as an array of 10 shorts (20 bytes) even though it is a single short.
```
jpetrie: These changes should be marked with a CHANGED FROM ORIGINAL header.

jpetrie (changes requested):

src/realmz_orig/save-direction-order.c
afkelsall: This block is already inside the pre-existing danapplegate `/* *** CHANGED FROM ORIGINAL IMPLEMENTATION *** ... */` marker that opens a few lines above (just off the top of this diff hunk) and closes with `/* *** END CHANGES *** */` right after the `fwrite`. I extended its comment to explain the 10-short retail layout. Happy to restructure if you would prefer the marker sit closer to the changed code.

jpetrie: No, that's fine, I just missed it being inside another block.

jpetrie (approved):

PR #242, opened 2026-06-09
menucontroller gets automatic memory management back

chromancer:
issue: general latency climbs as game plays on Mac. when Josh pointed out that #235 needed more draws, I instrumented it and found it grew even when #-of-syncs didn't.

cause: menu-driven again, it's #209. Xcode defaults ARC on, cmake defaults MRC. with no release, sync time climbs from 4ms each to 21ms each in ~3m of starting CoB.

fix:

* turn on ARC for the file (-fobjc-arc)
* add one __bridge for ARC

effects: time per sync is now 4ms flat. tested on Tahoe 26.5.1

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -242,6 +242,7 @@ set(realmz_lib_LINK_LIBRARIES
 
 if(APPLE)
     list(APPEND SOURCES src/macos/MenuController.mm)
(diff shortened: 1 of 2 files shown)


jpetrie (approved):
As we add more Objective-C or Objective-C++ files, we should make sure to handle them more generically.

Issue #112, opened 2025-08-05
Implement Scroll Bar Control


danapplegate:
Necessary for the trade/shop screen. See [TODO in WindowManager](https://github.com/danapplegate/realmz/blob/35daadc3b4db48154000df6b866c8a46d469f51d/src/WindowManager.cpp#L336-L337).

jpetrie:
#243 closes this.

PR #243, opened 2026-06-12
Fix shop scrollbars and live thumb drag

afkelsall:
The shop scrollbars did not render and the thumb could not be dragged. The track was painted with erase_rect, which honored the shop window stone background pattern instead of the scrollbar colors, the arrows were solid black squares, the thumb was a hollow outline, and TrackControl returned without changing the value.

Repaint the bar with foreground fills so it no longer depends on the installed pattern, reproducing the classic platinum scrollbar (recessed track, periwinkle thumb with grip, two arrow buttons grouped at the bottom). TrackControl now tracks the press: arrows draw pushed in and return their part only on release over the button, and the thumb drags live with the control value following the cursor. The hit-test and slider metrics match the draw code so clicks line up with what is shown.

Additionally, have implemented click highlighting for both the slider and arrow buttons, again identical to original.

QOL Enhancement: A second commit makes the item list follow the thumb during the drag. TrackControl calls a caller-supplied action proc whenever the value changes, the same way the Control Manager calls one for the arrows, and the shop passes a small proc that repaints the list. It fires only when the row value changes, so the redraw is throttled to one repaint per row. All of this lives in the SDL shim and shop logic, so macOS and Windows behave the same.

Relates to #112.

Port:
image

Original MacOS9 sheepshaver on left, port on right (can also tell from gamma difference):
image

I took both the above images from identical coordinates so I could beyond compare image diff them. The blue is diff within 38% tolerance due to gamma difference. Grey is no diff. Red would show up if diff outside tolerance.
image

### DIFF
--- a/src/WindowManager.cpp
+++ b/src/WindowManager.cpp
@@ -272,6 +272,10 @@ public:
   bool enabled;
   std::shared_ptr<Control> control; // May be null
 
+  // For scrollbars: which part is currently held down (0 = none). Set while
+  // TrackControl tracks a press so the held arrow can be drawn pushed in.
+  int16_t pressed_part = 0;
+
   static std::unordered_map<size_t, std::weak_ptr<DialogItem>> all_items;
 
   static std::shared_ptr<DialogItem> get_item_by_handle(size_t handle) {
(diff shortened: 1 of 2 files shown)


afkelsall:
I've just thought about this and there will be a slight issue with gamma. I've copied the scrollbars to be pixel perfect from the  original binary running in sheepshaver MacOS9. That has known gamma difference and if/when we fix the gamma in the port it will change the color of the scrollbars. We may need to get a gamma switch in and re-color the scrollbars.

fuzziqersoftware (approved):
Very nice work!

afkelsall:
The final commit is due to the above comment. The scrollbars need to match the gamma but I'd copied them from sheepshaver with different gamma. This is now corrected and they appear slightly darker and in line with the rest of the UI by default now. If the 2.59 gamma is selected, they match sheepshaver exactly now.

src/realmz_orig/choice.c
```diff
@@ -1,6 +1,19 @@
 #include "prototypes.h"
 #include "variables.h"
 
+/*********** live thumb-drag redraw ***********/
```
jpetrie, line 8: Changes here should get marked with a "changed from original" header.

src/WindowManager.cpp
```diff
@@ -509,42 +513,175 @@ struct DialogItem {
...
+        // because the shop window installs a stone background pixel pattern
+        // (BackPixPat) and erase_rect honors that pattern over the background
+        // color. The bevel layout was taken from the original 1994 build. Color
+        // values are stored pre-gamma (Mac OS 9 internal framebuffer level) so
+        // they render correctly when 2.59 color correction is active; without
+        // correction they match the raw Mac OS 9 output rather than the
+        // SheepShaver-presented output.
+        const RGBColor BLACK = {.red = 0x0000, .green = 0x0000, .blue = 0x0000};
+        const RGBColor TRK_FILL = {.red = 0xA9A9, .green = 0xA9A9, .blue = 0xA9A9};
+        const RGBColor TRK_S1 = {.red = 0x7777, .green = 0x7777, .blue = 0x7777}; // recessed shadow, darkest
```
jpetrie: A lot of these constant names are tough to follow, and I'm not sure they're referenced often/densely enough to warrant such concise names.

src/WindowManager.cpp
```diff
@@ -509,42 +513,175 @@ struct DialogItem {
...
+        const RGBColor AR_FACE = {.red = 0xDDDD, .green = 0xDDDD, .blue = 0xDDDD};
+        const RGBColor AR_HI = {.red = 0xFFFF, .green = 0xFFFF, .blue = 0xFFFF}; // arrow top-left highlight
+        const RGBColor AR_SH = {.red = 0xBABA, .green = 0xBABA, .blue = 0xBABA}; // arrow bottom-right shadow
+        // Pressed (pushed-in) arrow palette: darker face with the bevel inverted.
+        const RGBColor AR_FACE_P = {.red = 0x7777, .green = 0x7777, .blue = 0x7777};
+        const RGBColor AR_HI_P = {.red = 0x5555, .green = 0x5555, .blue = 0x5555}; // top-left, now dark
+        const RGBColor AR_SH_P = {.red = 0x9999, .green = 0x9999, .blue = 0x9999}; // bottom-right, now light
+
+        // Single pixel, horizontal run [x0..x1], and vertical run [y0..y1].
+        auto px = [&](int16_t x, int16_t y, const RGBColor& c) {
```
jpetrie: Why not make these actual functions?

src/WindowManager.cpp
```diff
@@ -509,42 +513,175 @@ struct DialogItem {
...
+        auto vspan = [&](int16_t x, int16_t y0, int16_t y1, const RGBColor& c) {
+          RGBForeColor(&c);
+          port.fill_rect(Rect{.top = y0, .left = x, .bottom = static_cast<int16_t>(y1 + 1), .right = static_cast<int16_t>(x + 1)});
+        };
+
+        const int16_t bx0 = r.left, bx1 = static_cast<int16_t>(r.right - 1); // black border columns
+        const int16_t by0 = r.top, by1 = static_cast<int16_t>(r.bottom - 1); // black border rows
+        const int16_t ix0 = static_cast<int16_t>(bx0 + 1), ix1 = static_cast<int16_t>(bx1 - 1); // interior columns
+        const int16_t th_h = w; // thumb border-to-border height
+        const int16_t ar_h = static_cast<int16_t>(w - 1); // each arrow button border-to-border height
```
jpetrie: I don't think the excessive `static_cast`ing here is needed. Here, for example, if we ensured `w` was an `int16_t` instead of...whatever it is due to `auto`... by casting it at the initialization, we wouldn't need to cast every time we use it after subtracting a constant. The casts make this harder to follow than I think it needs to be, and every time I see one I have to wonder if we're actually handing the edge cases where the input to the cast is out of range correctly.

src/WindowManager.cpp
```diff
@@ -1843,7 +1993,91 @@ short TrackControl(ControlHandle handle, Point pt, ProcPtr action_proc) {
...
+        int new_value = item->control->min + static_cast<int>(static_cast<double>(off) / slider_range * value_range + 0.5);
+        new_value = std::clamp<int>(new_value, item->control->min, item->control->max);
+        if (new_value != item->control->value) {
+          item->control->value = static_cast<int16_t>(new_value);
+          render_window_for_item(item);
+          if (have_live_proc) {
+            live_proc(handle, kControlIndicatorPart);
+          }
+        }
+        SDL_Delay(15);
```
jpetrie, line 2087: What happens if we don't throttle here?

jpetrie (changes requested):

src/WindowManager.cpp
afkelsall: Without it the drag loop becomes a tight busy-wait: `Button()`/`GetMouseGlobal` just read cached state, so the loop would spin a core at 100% polling the mouse. The redraw itself is already throttled by the `new_value != value` check (one repaint per row), so the delay isn't about redraw rate, it just caps the poll loop.

I haven't however tested this without it.

jpetrie (approved):

PR #244, opened 2026-06-12
Show the dragged item icon in the shop

afkelsall:
What this is:
- Restores the original game's visible item dragging in the shop. While buying or selling, the cursor hides and the item icon follows the mouse, matching the 1994 build.
- Previously the dragged icon was written into the in-memory screen buffer but never shown, so the cursor just disappeared with nothing in its place.

How it works:
- Adds a way to present the screen buffer to the window without recompositing the window stack, called as the icon moves. Recompositing would immediately erase the transient icon.
- Draws the icon from a saved pristine copy with a transparent blit, so it stays on top of any background instead of blending with it.
- Saves the pixels hidden under the icon in a separate buffer and restores them cleanly, which stops the icon smearing into duplicates when dragged over its own list slot.
- Clamps ShowCursor to the visible cursor level, as on Classic Mac OS. The game calls ShowCursor far more than HideCursor, which had driven the level negative and left the cursor visible during subsequent drags.

Stays macOS and Windows compatible: the change lives in the SDL screen presentation path and the shared QuickDraw cursor and CopyBits shims, with no platform specific code.

Both MacOS9 and port blend the icon over other colors:
image

This is now fixed in this PR too:
image

Dragging morning star from shop:
image

### DIFF
--- a/src/realmz_orig/moveicon.c
+++ b/src/realmz_orig/moveicon.c
@@ -27,7 +27,7 @@ void itemcost(short mode) /****** 1 = sell, else buy ****/
 void moveicon(void) {
   float charge1, charge2 = 0;
   short valu, max, tempshop;
-  Rect iconpict, iconnp, iconop, iconstore, store;
+  Rect iconpict, iconnp, iconop, iconstore, store, storebg;
   Point oldpoint;
   Boolean empty = 0;
   Boolean doupdateshop = 0;
@@ -36,6 +36,7 @@ void moveicon(void) {
   short leftvalue, temp;
   struct itemattr tempitem;
   Boolean baditem = 0;
+  Boolean haveold = 0;
   int index;
   /* *** CHANGED FROM ORIGINAL IMPLEMENTATION ***
    * NOTE(danapplegate): It seems that the global screen buffer was directly accessible. We've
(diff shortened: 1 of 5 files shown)


fuzziqersoftware (approved):

src/realmz_orig/moveicon.c
```diff
@@ -27,7 +27,7 @@ void itemcost(short mode) /****** 1 = sell, else buy ****/
 void moveicon(void) {
   float charge1, charge2 = 0;
   short valu, max, tempshop;
-  Rect iconpict, iconnp, iconop, iconstore, store;
+  Rect iconpict, iconnp, iconop, iconstore, store, storebg;
```
jpetrie: These changes should be marked with a "changed from original" comment such as [this one](https://github.com/Realmz-Castle/realmz/blob/fc143ecb7d54b1f7be3ff7e714fea450297b8bb9/src/realmz_orig/warn.c#L61).

jpetrie (dismissed):

src/realmz_orig/moveicon.c
afkelsall: Sorted.

danapplegate (approved):

PR #247, opened 2026-06-14
Reuse one streaming texture when presenting

afkelsall:
This change came from the performance investigation I did, albeit on the Debug build. This was one of two small things to come out of it. The impact of this on the actual render was only measured to be a few ms on the Release build but figured I'd still put it up to consider/discuss, but happy to decide not to merge this and close if not considered necessary.

## Why
- The compositor created and freed a new SDL surface and GPU texture on every present, so each frame that updated the screen paid a GPU texture allocation and upload it did not need.
- This affected every redraw, including animations.

## How
- Keep one streaming texture, created once and recreated only when the window size changes, and update it in place with SDL_UpdateTexture.
- Recreate the texture if an update fails, so the window recovers after a render device reset.
- Output is unchanged: same pixel format, same full frame upload, same default scale mode.

### DIFF
--- a/src/WindowManager.cpp
+++ b/src/WindowManager.cpp
@@ -1178,19 +1178,30 @@ void WindowManager::recomposite(std::shared_ptr<Window> updated_window) {
         wm_log.info_f("Writing debug{}.bmp", debug_number);
         phosg::save_file(std::format("debug{}.bmp", debug_number++), this->screen_port.data.serialize(phosg::ImageFormat::WINDOWS_BITMAP));
       }
-      // Reuse one streaming texture across frames instead of creating a new
-      // surface and texture every present.
-      if (!this->screen_texture || this->screen_texture_w != w || this->screen_texture_h != h) {
-        this->screen_texture = sdl_make_unique(SDL_CreateTexture(
-            renderer, SDL_PIXELFORMAT_RGBA8888, SDL_TEXTUREACCESS_STREAMING, w, h));
-        this->screen_texture_w = w;
-        this->screen_texture_h = h;
+      // Reuse one streaming texture instead of creating a new surface and
+      // texture every present; recreate it on size change or update failure.
+      auto ensure_texture = [&] {
+        if (!this->screen_texture || this->screen_texture_w != w || this->screen_texture_h != h) {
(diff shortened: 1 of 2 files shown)


src/WindowManager.cpp
```diff
@@ -1172,23 +1172,25 @@ void WindowManager::recomposite(std::shared_ptr<Window> updated_window) {
       wm_log.error_f("Could not get window renderer: {}", SDL_GetError());
     } else {
 
-      auto w = this->screen_port.data.get_width();
-      auto h = this->screen_port.data.get_height();
+      int w = static_cast<int>(this->screen_port.data.get_width());
```
jpetrie, line 1175: I can understand wanting to be more explicit than `auto` with the type, but why change the actual type used (from `size_t`) and add a cast? What happens if the width can't fit into an `int` in some extremely-high-resolution future?

src/WindowManager.cpp
```diff
@@ -1178,19 +1178,30 @@ void WindowManager::recomposite(std::shared_ptr<Window> updated_window) {
...
-      // Reuse one streaming texture across frames instead of creating a new
-      // surface and texture every present.
-      if (!this->screen_texture || this->screen_texture_w != w || this->screen_texture_h != h) {
-        this->screen_texture = sdl_make_unique(SDL_CreateTexture(
-            renderer, SDL_PIXELFORMAT_RGBA8888, SDL_TEXTUREACCESS_STREAMING, w, h));
-        this->screen_texture_w = w;
-        this->screen_texture_h = h;
+      // Reuse one streaming texture instead of creating a new surface and
+      // texture every present; recreate it on size change or update failure.
+      auto ensure_texture = [&] {
```
jpetrie, line 1183: Why a lambda versus a member function? Somewhat related, it looks like we're trying to do just-in-time creation of the texture here, either initially or during any kind of property change. Could we instead create the texture during construction, and re-create in response to SDL's events indicating a window size change?

jpetrie (changes requested):

afkelsall:
Given the game renders fine in release mode, I'm inclined to not make unnecessary changes after thinking about this for a few days.

PR #248, opened 2026-06-14
Speed up and fix the post-fight treasure screen

afkelsall:
## Why
- Selecting an item on the treasure screen was slow and the pickup sparkle dragged, because every QuickDraw call recomposites and presents the whole screen. Albeit this was on Debug build and the Release build was a non-issue, the efficiency may still be desired.
- Hovering an item took a noticeable moment and briefly blocked clicks, and grabbing items in quick succession felt laggy.
- The sparkle left stray coloured pixels, the right edge of the grid kept fragments, and a text blip sound replayed when cursor exits the white item area.

## How
- Batch the item-keep redraw, the hover redraw, and the sparkle loop so each presents once instead of once per drawing call, and play the sparkle as a quick flourish.
- Skip the pickup flourish, and the idle delay, when another click is already queued, so rapid selection stays responsive.
- Repaint the emptied cell and its neighbours to clear sparkle leftovers, widen the cleared region at the right edge, and preserve the inbooty flag so the hover preview does not replay the text blip.


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## Unreleased
```
jpetrie, line 1: IMO we shouldn't track unreleased features here:
- Assuming we are using good commit messages, GitHub can reconstruct this list for us when preparing a new release. 
- It increases contention for the file, leading to more merges and potential merge conflicts.
- It's easy to forget to add stuff, and even easier to forget to _remove_ stuff if we have to walk back a change.

I realize a lot of these changes apply to the very idea of this file at all, but it already exists so in this context I'd just like to advocate for not making it a source of friction.

src/realmz_orig/booty.c
```diff
@@ -1600,6 +1600,7 @@ void booty(short mode) {
                       temprect = icon;
                       InsetRect(&temprect, 10, 10);
 
+                      int enable_recomposite = WindowManager_SetEnableRecomposite(0);
```
jpetrie, line 1603: These changes would need to be flagged as modifications to the original source per our contribution guidelines in the README.

jpetrie (changes requested):

src/realmz_orig/main.c
```diff
@@ -2161,6 +2162,16 @@ void quickinfo(int who, int itemnumber, int itemid, int where) {
       break;
   }
 
+  /* *** CHANGED FROM ORIGINAL IMPLEMENTATION ***
+   * quickinfo is also called from the booty (treasure) screen in the
+   * large-window layout while an item is being hovered, and there inbooty must
+   * stay TRUE so that textbox keeps quiet. The original code unconditionally
+   * cleared inbooty after drawing the description below, which let the text
+   * blip sound fire every time the cursor afterward left the treasure area.
+   * Save the prior value and restore it instead of forcing FALSE. */
```
jpetrie, line 2171: So this was a bug in the original, as well?

src/realmz_orig/booty.c
```diff
@@ -714,13 +714,20 @@ void booty(short mode) {
 backup:
 
   PenSize(2, 2);
+  a = 0;
```
jpetrie, line 717: This variable name isn't very descriptive. A quick look shows it's maybe only used in the original code for holding the GetNextEvent() response? I can see why it might be tempting to reuse it, but it makes the change here less clear. Whatever state we're tracking here should probably get its own variable with a descriptive name.

jpetrie:
I left a few comments on individual commits, but I'm not convinced this is a set of changes we need to make, given that release mode (and thus release build) performance is acceptable. 

It is annoying that debug builds have a bit of a performance issue here, but I don't think it's unreasonable and I don't think it warrants the amount of changes to the original needed.

afkelsall:
I think given the release mode issues I'll close this PR and if we decide to fix the item selection sparkle or find other display issues, we can cherry pick from this or use it as a base to identify the issue and build something more targeted. There's a lot in here that's just not needed and keeping changes to a minimum is probably desirable.

afkelsall:
I've played through a few battles in the arena and this PR adds a lot of QOL improvements outside the pure redraw debug build enhancements that were in there. The sparle leaves a of residue on every item. Half the items don't take because you've clicked too fast. THere's a boop as you leave the item area. I"m reinstating this branch locally to test, but I"m inclined to discuss this further.

PR #251, opened 2026-06-15
Use the original Black Chancery font and baseline-anchor pen text

afkelsall:
The stat numbers drawn in Black Chancery used a different font than the original and sat slightly off, with old-style descenders running past the boxes they were drawn into.

Why:
- The bundled Black Chancery was a different digitization than the one the game embeds, with a taller line height and deeper descenders, so numbers rendered larger and lower than the original.
- Pen-based text (MoveTo then DrawText) was centered in its display rect instead of placed on the baseline, so it sat a pixel or so off from where the original drew it.

How:
- Replace resources/Black Chancery.ttf with the public-domain Black Chancery that ships inside the game data (the autotraced version the original renders), so glyph shapes and vertical metrics match.
- Draw pen-based text on its baseline (surface top at baseline minus ascent) rather than centering it. Dialog item text keeps the existing centering, so only MoveTo-drawn text changes.
- Compatible with macOS and Windows since it only touches the shared shim and a bundled font.

Supersedes #249, which clipped the roster numbers to their boxes; this fixes the cause instead by matching the original font's metrics and placement.

Black chancery font original vs bundled with port comparison. This is the cause of the redraw artifacts in #249 before I realized the fonts were just _different_. Thanks to everyone in discord helping identify that, I was then able to physically extract the original font from the actual "The Family Jewels.rsrc" file from the original game and create a ttf font from it for this PR. Below image is the various renderings of each version compared and here you can see why we were seeing rendering issues, the top and bottom tail extensions were far longer than the original version.
compare_digits

Character list on mains screen: Original on left, port on right.
image

Item screen with character stats: Original on left port on right.
image

You can see the diffs are not _perfect_ and I'm holding myself back getting it perfect here, although I tried.

For comparison to the above shots, this was the previous state of the port:
character-list-font-original-problem


src/QuickDraw.cpp
```diff
@@ -174,7 +174,7 @@ phosg::ImageRGBA8888N image_for_sdl_surface(SDL_Surface* surface) {
   return ret;
 }
 
-bool CCGrafPort::draw_text_ttf(TTF_Font* font, const std::string& processed_text, const Rect& rect) {
+bool CCGrafPort::draw_text_ttf(TTF_Font* font, const std::string& processed_text, const Rect& rect, bool baseline_anchor) {
```
jpetrie: From the commit message, it sounds like this change is fixing a general bug with text drawing, so what utility is this flag providing? When would we want to pass false to get the old behavior?

src/QuickDraw.cpp
```diff
@@ -185,10 +185,25 @@ bool CCGrafPort::draw_text_ttf(TTF_Font* font, const std::string& processed_text
     return false;
   } else {
     auto img = image_for_sdl_surface(text_surface.get());
-    // This is annoying, but it seems there isn't a better way to do it... if the rendered text height exceeds the
-    // target rect, we trim off some of the top rows to center it vertically. This isn't exactly correct (some text
-    // appears to be off by 1 or 2 pixels sometimes) but it will do for now. There aren't good metrics provided by
-    // SDL_ttf for this (ascent/height don't match the actual amount we need to trim) so we have to do this instead.
+    if (baseline_anchor) {
+      // Pen-based DrawText places the pen at the text baseline, and QuickDraw draws glyphs
```
jpetrie: I don't think we actually call or even implement DrawText, so I'm not sure I follow this comment.

When you talk about the caller building the rectangle so that bottom + descent equal's the pen's vertical position, I'm not seeing how exactly the corresponds to most of the initial call stacks of this function when you e.g., start a new game, which are actually called from the Dialog Manager implementions (DrawDialog, SetDialogItemText, et cetera).

I also note that the original comment in this function noted that SDL's ascent/descent metrics weren't sufficient for our needs, but we're using them here and I'm not sure that the original comments concern has been resolved?

jpetrie (changes requested):

src/QuickDraw.cpp
afkelsall: The flag was really just telling the renderer which caller it had, which wasn't great. Have changed it so `draw_text_ttf` is now `render_text_ttf`, which only renders the text to an image and returns it. Each caller does its own placement.

* `draw_text(txt)` anchors the surface on the baseline
* `draw_text(text, rect)` keeps the existing centering.
 
To answer the original question directly, the box path can't be baseline-anchored because those rects are Dialog Manager item boxes (no pen/baseline in them and often multi-line). Top/baseline-anchoring them bunches up multi-line labels like on the party-selection screen. So the two paths genuinely need different placement, and that now lives in the callers rather than behind a mode flag. 

I saw these problems emerge in my testing before realising the need to support both options.

src/QuickDraw.cpp
afkelsall: Have reworded this

1. Dropped the "DrawText" phrasing. There's no such call because the path is the pen overload `draw text(text)` reached via `MoveTo` + `DrawString`.
2. This baseline placement is only on the pen overload. The Dialog Manager calls you see when starting a game (DrawDialog/SetDialogItemText) go through `draw_text(text, rect)` which still centers and is untouched. None of those call stacks hit the baseline math
3. The old "metrics aren't sufficient" note was about the centering/trim case, where ascent/height don't tell you how many rows to trim to fit a box. Baseline anchoring uses ascent for a different purpose: SDL_ttf places the rendered surface's baseline exactly `TTF_GetFontAscent` rows below its top, so aligning that to the pen baseline is exact rather than approximate.

I left the original caveat on the centering branch, since that concern still stands there.

jpetrie (approved):

PR #252, opened 2026-06-21
Fix Open Lock spell chance

iSynic:
## Summary

Open Lock encounter spells compute their success chance from `thief.sound[1] * powerlevel`, but the final success roll used `thief.sound[2] * powerlevel` instead. `sound[2]` is used by the Destroy Trap path, so Open Lock could succeed or fail based on the trap-disarm spell chance instead of its own chance.

This changes the Open Lock roll to use the already-computed `knockchance` value.

## Verification

- Inspected the Destroy Trap and Open Lock encounter spell handling in `src/realmz_orig/encounters.c`.
- Ran `git diff --check` locally. It reported only the existing CRLF warning for the touched file.
- Built `Realmz` successfully on Windows with CMake/Ninja using MSYS2 clang.

### DIFF
--- a/src/realmz_orig/encounters.c
+++ b/src/realmz_orig/encounters.c
@@ -977,7 +977,9 @@ tryagain:
             thief.type[6] = TRUE;
           }
 
-          /*** knock uses sound[1]; sound[2] is disarmtrap ***/
+          /* *** CHANGED FROM ORIGINAL IMPLEMENTATION ***
+           * Open Lock uses thief.sound[1]; thief.sound[2] is Destroy Trap.
+           */
           if (Rand(100) < knockchance) {
             thief.type[9] = FALSE; /*** success ***/
 
--- a/src/realmz_orig/encounters.c
+++ b/src/realmz_orig/encounters.c
@@ -977,7 +977,8 @@ tryagain:
             thief.type[6] = TRUE;
           }
 
-          if (Rand(100) < thief.sound[2] * powerlevel) {
+          /*** knock uses sound[1]; sound[2] is disarmtrap ***/
+          if (Rand(100) < knockchance) {
             thief.type[9] = FALSE; /*** success ***/
 
             sound(thief.sounds[6]);


src/realmz_orig/encounters.c
```diff
@@ -977,7 +977,8 @@ short encounter2(void) {
             thief.type[6] = TRUE;
           }
 
-          if (Rand(100) < thief.sound[2] * powerlevel) {
+          /*** knock uses sound[1]; sound[2] is disarmtrap ***/
```
jpetrie: This should be marked with a "changed from original" comment using the same format as, for example, [this one](https://github.com/Realmz-Castle/realmz/blob/fc143ecb7d54b1f7be3ff7e714fea450297b8bb9/src/realmz_orig/warn.c#L61).

jpetrie (changes requested):

src/realmz_orig/encounters.c
iSynic: Sorry about that, my first PR.

Added the standard “CHANGED FROM ORIGINAL IMPLEMENTATION” marker above the Open Lock change.

jpetrie:
I was a little hung up on understanding what exactly these `sound[n]` fields mean - sometimes, like in this code, they are used as scalars in some spell power or chance calculation, and other times they are passed to the `sound()` function. 

When I look at `resource_dasm`s map of the [rogue encounter data](https://github.com/fuzziqersoftware/resource_dasm/blob/000f7774762a68dd810096fe9baef7a40ca35179/src/RealmzScenarioData.hh#L502), however, it makes it a bit clearer: the last three 16-bit values (which correspond to the `thief.sound[]` array) are "prompt sound", "% per level to open" and "% per level to disable."

`sound[0]` is what is passed to `sound()`, and the other two are used as described. The use of `sound[2]` in the original check, addressed by this PR, does appear to have been a copy-paste style of error (the disarm trap block is above this one).

jpetrie (approved):

PR #253, opened 2026-06-21
Honor synchronous SndPlay requests

iSynic:
## Summary

Realmz uses negative sound ids to request synchronous playback. The `sound()` wrapper passes those through to `SndPlay(..., FALSE)`, while positive ids use `SndPlay(..., TRUE)`.

`SndPlay` accepted the `async` argument but ignored it, so synchronous sound requests queued audio and returned immediately. This keeps async playback unchanged. For `async == FALSE`, `play_sound` queues the decoded sound, flushes the SDL stream, and waits for the stream to drain before returning.

### DIFF
--- a/src/SoundManager.cpp
+++ b/src/SoundManager.cpp
@@ -15,8 +15,6 @@
 static phosg::PrefixedLogger sm_log("[SoundManager] ");
 
 constexpr size_t OUTPUT_SAMPLE_RATE = 48000;
-constexpr uint32_t SYNC_SOUND_GRACE_MS = 100;
-constexpr uint32_t SYNC_SOUND_MAX_WAIT_MS = 5000;
 
 class SoundManager {
 public:
@@ -74,13 +72,13 @@ public:
     return channel;
   }
 
-  bool play_sound(SDL_AudioStream* sdlAudioStream, Handle data_handle, uint32_t* duration_ms) {
+  void play_sound(SDL_AudioStream* sdlAudioStream, Handle data_handle, bool async) {
     std::shared_ptr<const Sound> sound;
     try {
       sound = this->sound_for_handle(data_handle);
     } catch (const std::exception& e) {
       sm_log.warning_f("Can't find or decode sound: {}", e.what());
-      return false;
+      return;
     }
 
     // SDL_PutAudioStreamData takes an int to represent the audio stream size in bytes. It's unlikely that any sound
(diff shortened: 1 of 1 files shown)


src/SoundManager.cpp
```diff
@@ -72,25 +74,67 @@ class SoundManager {
     return channel;
   }
 
-  void play_sound(SDL_AudioStream* sdlAudioStream, Handle data_handle) {
+  bool play_sound(SDL_AudioStream* sdlAudioStream, Handle data_handle, uint32_t* duration_ms) {
```
fuzziqersoftware: This is an odd API design, since we'll never want to call `wait_for_playback` independently of `play_sound`. Instead, we should just pass `bool async` to this function and the wait should occur within this function after writing to the SDL stream.

src/SoundManager.cpp
```diff
@@ -72,25 +74,67 @@ class SoundManager {
...
+  void wait_for_playback(SDL_AudioStream* sdlAudioStream, uint32_t duration_ms) {
+    if (!SDL_FlushAudioStream(sdlAudioStream)) {
+      sm_log.warning_f("Could not flush audio stream: {}", SDL_GetError());
+      return;
+    }
+
+    const uint64_t timeout_ms = std::min<uint64_t>(
+        static_cast<uint64_t>(duration_ms) + SYNC_SOUND_GRACE_MS,
+        SYNC_SOUND_MAX_WAIT_MS);
+    const uint64_t deadline = SDL_GetTicks() + timeout_ms;
```
fuzziqersoftware: Why don't we just use the actual duration of the sound, or just wait until the channel is idle? Seems like SndPlay would have done one of those two things on Classic Mac OS; I doubt it had a deadline parameter.

fuzziqersoftware (changes requested):

src/SoundManager.cpp
```diff
@@ -115,6 +124,19 @@ class SoundManager {
...
+      if ((bytes_per_sample == 0) || (this->num_channels == 0) || (this->sample_rate == 0)) {
+        return 0;
+      }
+
+      const size_t bytes_per_frame = bytes_per_sample * this->num_channels;
+      const size_t num_frames = this->data.size() / bytes_per_frame;
+      return static_cast<uint32_t>(
+          (static_cast<uint64_t>(num_frames) * 1000 + this->sample_rate - 1) /
+          this->sample_rate);
+    }
```
fuzziqersoftware: We shouldn't need this at all, right? If we use something like this above then it shouldn't be necessary:

    while (SDL_GetAudioStreamAvailable(stream) > 0) {
        SDL_Delay(10);
    }

iSynic: That makes sense. I’ll simplify it to wait on the stream after flushing and remove the duration helper.

Edit: I believe the requested changes are addressed now. The sync wait is folded into play_sound(..., async), and the separate duration/deadline helper was removed.

fuzziqersoftware (approved):

PR #256, opened 2026-06-22
Port menu

chromancer:
Mac and Windows native Port menu @afkelsall with shared base presentation options. I have been test driving this for a while in concert with him as I'm sure most have seen in discord. I believe it to be entirely stable. I could continue polishing it but I think it probably best to get this out there and community test driving it for feedback.

features:
* 800x600 as logical size; this size is default.
* Resizable window supports Mac native fullscreen
* Filter submenu (pixel art / linear / nearest)
* Scaling shortcuts by half step (I prefer integer but ymmv)
* Option to lock window to 4:3
* Option to color correct according to several gamma styles.
* user window size, position, aspect-lock, and gamma style persisted via JSON port prefs

testing: built on Mac, menu/filters/scaling/gamma/persistence all work as expected. tested by @afkelsall on Windows side.

known issues: when leaving Mac native fullscreen, intermittently, the window unexpectedly moves.


src/macos/MenuController.mm
```diff
@@ -108,6 +109,11 @@ - (IBAction)MCHandleMenuClick:(id)sender {
   callback([identifier menuID], [identifier itemID]);
 }
 
+- (IBAction)MCHandlePortItem:(id)sender {
+  NSInteger tag = [sender tag];
+  PortMenu_Apply((PortCmdKind)(tag / 256), (int)(tag % 256));
```
jpetrie: I'd like to see utility functions that are encapsulating the intent behind the divisions by 256 (such as `getPortCmd()` and `getPortCmdIndex()`).

But stepping back a bit, why are we encoding the tag in this fashion at all versus just assigning each leaf node menu item a value in the command enum?

src/macos/MenuController.mm
```diff
@@ -122,6 +128,80 @@ - (void)MCCreateMenu:(const MenuList&)menuList {
...
+  [portMenu addItem:scaleItem];
+  NSMenu* scaleMenu = [[NSMenu alloc] initWithTitle:@"Scale"];
+  [scaleMenu setAutoenablesItems:NO];
+  scaleMenu.delegate = self;
+  for (int i = 0; i < kPortScaleCount; i++) {
+    NSMenuItem* item = [scaleMenu addItemWithTitle:[NSString stringWithUTF8String:kPortScales[i].title]
+                                            action:@selector(MCHandlePortItem:)
+                                     keyEquivalent:@""];
+    [item setTarget:self];
+    [item setTag:(NSInteger)(PortCmdScale * 256 + i)];
```
jpetrie: As with line 114, it would be nice to have a function that self-documents that the math here is encoding a packed port menu command tag.

src/macos/MenuController.mm
```diff
@@ -122,6 +128,80 @@ - (void)MCCreateMenu:(const MenuList&)menuList {
...
+    NSMenuItem* item = [gammaMenu addItemWithTitle:[NSString stringWithUTF8String:kPortGammaOptions[i].title]
+                                           action:@selector(MCHandlePortItem:)
+                                    keyEquivalent:@""];
+    [item setTarget:self];
+    [item setTag:(NSInteger)(PortCmdGamma * 256 + i)];
+  }
+  [portMenu setSubmenu:gammaMenu forItem:gammaItem];
+
+  [_menuObject setSubmenu:portMenu forItem:portItem];
+}
```
jpetrie, line 191: It might be useful to have a menu item to restore the window size to default, in case somebody accidentally drags it to a weird size.

Doesn't need to be addressed for this review though.

src/macos/WindowAspect.mm
```diff
@@ -0,0 +1,17 @@
+#include "WindowAspect.h"
+
+#import <Cocoa/Cocoa.h>
+#include <SDL3/SDL_properties.h>
+#include <SDL3/SDL_video.h>
+
+// Clears the 4:3 lock via contentResizeIncrements, which resets contentAspectRatio
```
jpetrie: I don't see this documented [here](https://developer.apple.com/documentation/appkit/nswindow/contentresizeincrements?language=objc), and why not just set `contentAspectRatio` directly anyhow?

src/windows/WinMenuController.cpp
```diff
@@ -4,11 +4,109 @@
...
+#include "../PortMenu.hpp"
 #include "./WinMenuController.hpp"
 #include <utility>
 
 static phosg::PrefixedLogger wmc_log("[WinMenuController] ");
 
+// Command IDs for the Port menu. They sit in a reserved high range so they never
+// collide with the packed (menu_id, item_id) identifiers the game's own menus use,
+// which top out at 0x7F7F because menu_id and item_id are each a single byte.
+static constexpr WORD PORT_FILTER_BASE = 0xE000; // + filter index
```
jpetrie: Why do we have to encoded the command IDs separately for macOS and Windows like this? I'd rather these at least reference a shared enumeration or list of constants to avoid having to remember to update the item in two places.

src/EventManager.cpp
```diff
@@ -413,13 +419,26 @@ class EventManager {
     em_log.debug_f("Enqueued event (what={}, message=0x{:08X}, when=0x{:08X}, where=(h={}, v={}), modifiers=0x{:04X})", name_for_event_type(ev.what), ev.message, ev.when, ev.where.h, ev.where.v, ev.modifiers);
   }
 
-  void enqueue_sdl_event(const SDL_Event& e) {
+  void enqueue_sdl_event(const SDL_Event& e_in) {
+    SDL_Event e = e_in;
```
jpetrie: If we're copying `e_in` anyway, can we not just change the signature to `enqueue_sdl_event(SDL_Event e)`?

src/PortMenu.hpp
```diff
@@ -0,0 +1,67 @@
+#pragma once
+
+#include <SDL3/SDL_surface.h>
+
+inline constexpr int kLogicalWidth = 800;
```
jpetrie: This could mean the width/height of anything if referenced out of context. `kLogicalWindowWidth` or similar might be better names.

src/PortMenu.hpp
```diff
@@ -0,0 +1,67 @@
...
+struct PortGammaOption {
+  const char* title;
+  float display_gamma;
+};
+
+inline constexpr PortGammaOption kPortGammaOptions[] = {
+    {"Off",  0.0f},
+    {"2.0",  2.0f},
+    {"2.2",  2.2f},
+    {"2.59 (SheepShaver)", 2.59f},
```
jpetrie, line 54: Documenting this as the one SheepShaver uses is helpful - what are the others?

src/PortPrefs.cpp
```diff
@@ -0,0 +1,135 @@
...
+    return SDL_SCALEMODE_LINEAR;
+  } else if (name == "nearest") {
+    return SDL_SCALEMODE_NEAREST;
+  }
+  return SDL_SCALEMODE_PIXELART;
+}
+
+static double gamma_value_for_idx(int idx) {
+  if (idx < 0 || idx >= kPortGammaCount) {
+    return kPortGammaOptions[0].display_gamma;
```
jpetrie: This state is indicative of a programmer's error somewhere - might be better to log or assert rather than fix the bad input.

src/PortPrefs.cpp
```diff
@@ -0,0 +1,135 @@
...
+
+  try {
+    auto root = phosg::JSON::parse(data);
+    prefs.window_w = std::clamp(static_cast<int>(root.get_int("window_w", prefs.window_w)), MIN_W, MAX_W);
+    prefs.window_h = std::clamp(static_cast<int>(root.get_int("window_h", prefs.window_h)), MIN_H, MAX_H);
+    prefs.window_x = static_cast<int>(root.get_int("window_x", prefs.window_x));
+    prefs.window_y = static_cast<int>(root.get_int("window_y", prefs.window_y));
+    prefs.scale_mode = scale_mode_for_name(root.get_string("filter", name_for_scale_mode(prefs.scale_mode)));
+    prefs.aspect_locked = root.get_bool("aspect_locked", prefs.aspect_locked);
+    prefs.gamma_idx = gamma_idx_for_value(root.get_float("gamma", gamma_value_for_idx(prefs.gamma_idx)));
```
jpetrie, line 99: This makes me a bit wary, since we're doing a floating-point equality check on a deserialized value. Can we guarantee it will always round-trip exactly? Storing a more stable value (fixed point gamma? index?) or doing a search for the nearest matching gamma would be more robust.

src/WindowManager.hpp
```diff
@@ -101,6 +101,19 @@ class WindowManager {
   sdl_window_shared sdl_window;
   bool text_editing_active = false;
   bool recomposite_enabled = true;
+  SDL_ScaleMode scale_mode = SDL_SCALEMODE_PIXELART;
+  bool aspect_locked = true;
+  int gamma_idx = 0;
+  int windowed_w = 800;
```
jpetrie: We should use the constants for window width/height here, no?

jpetrie (changes requested):
Because the port menu is hard-coded (which I think is fine), it seems a bit over-complicated to then pack a command family and specific command into an integer to store in the tag so that we can provide a generic selector in the form of `MCHandlePortItem:`. 

Why not just bind to specific methods, or methods for filter/scale/etc that take a block or use the tag to pass the specific scaling factor they need? It seems like the current approach creates a bit more overhead to adding new items given the number of places enums or cases need to be updated to match.

src/WindowManager.hpp
chromancer: Yup missed this one.

src/PortMenu.hpp
chromancer: There was a whole discussion about this in the discord that I actually don't recall. @afkelsall can you comment?

src/macos/MenuController.mm
chromancer: I agree and considered it but it would duplicate Scaling > 1x. Since you're on the same track I think making this obvious to the user would be good

Shall I rename Scaling > 1x to "1x (Default)" or "1x (Original)"? Either signposts nicely

src/PortMenu.hpp
chromancer: Sure! done

src/macos/MenuController.mm
jpetrie: Does that set the window size back to its default as well? I thought I tried that and didn't notice anything, but maybe I just didn't pay enough attention.

I like "1x (Default)" better than "original," though, for what it's worth.

src/macos/WindowAspect.mm
chromancer: sorry, this is a _garbage_ comment. I let the llm review what I was doing and it snuck this rewrite of the comment into my uncommitted work. I'll rewrite the comment here

We _should_ set ratio directly the sane documented way. But we can't because the sane documented way crashes immediately on the next native window touch of any kind, resize/fullscreen, you name it. I thought this was SDL but followed it up to #14229, it's something in AppKit.

the workaround is just a result of disjoint state. you can't both lock the aspect ratio to 4:3 and resize in 1,1 increments, so this clears the lock and lets us move on.

src/PortMenu.hpp
afkelsall: The others just provide somewhat of a gradient to "off" or no gamma change. They were there for me to see the impact of different values and felt some people may find off too dark and sheepshaver too bright. I was keen to replicate the appearance of the original game when played on an actual Mac. The port does appear too dark to my eyes. I'm guessing a little here, but felt if we're putting in gamma settings, some people may want to choose something in between. That was the extent of my thinking here :)

src/macos/MenuController.mm
chromancer: yeah it uses SDL to just set the window… I can't think of a concrete way it wouldn't without some other interaction

src/PortPrefs.cpp
chromancer: search for nearest seems fine. this worked, but who knows if it will continue to is a decent point

src/PortPrefs.cpp
chromancer: assert, fair enough

src/EventManager.cpp
chromancer: a blunder! done

src/macos/MenuController.mm
chromancer: "why encode at all?" i think is an excellent question and one I believe is the best answer, running this back and forth with @afkelsall right now

chromancer: OK, after some discussion, we agree your overall suggestion is best. do away with this in favor of a flat list with shared id constants, windows handles this fine with base offset.

chromancer:
OK, think that's everything addressed for now. @afkelsall has has rebuilt on windows, looks good. lmk anything else you have for me

jpetrie (approved):

PR #258, opened 2026-06-22
stop two more crashes in party select

chromancer:
there are two more active crashes in party select from my original issue, one I mentioned and one I didn't see until recently.

1. if an oversize (non-character) file exists and is added to party. garbage is parsed as a character and crashes downstream.
2. if a subdirectory exists. mac_fopen returns null for directories (correct behavior for emulating classic). but passing a null FILE* to fread is immediate segfault.

the commit adds two add'l guards for the above.

my principle is that while these result from things that should not happen (there's no intended reason to have directories in Character Files, nor is there any good reason to add a non-character file to your party) we should guard an inexperienced user against crashes by eliding junk files completely and skipping directories

### DIFF
--- a/src/realmz_orig/partyselect-helpers.cpp
+++ b/src/realmz_orig/partyselect-helpers.cpp
@@ -32,7 +32,15 @@ extern "C" void update_character_files_list() {
   for (const auto& filename : mac_list_directory(":Character Files")) {
     try {
       auto f = mac_fopen_unique(std::format(":Character Files:{}", filename), "rb");
+      if (!f) {
+        characters_log.info_f("Skipped {} (could not be opened).", filename);
+        continue;
+      }
       auto ch = phosg::freadx<struct character>(f.get());
+      if (!phosg::fread(f.get(), 1).empty()) {
+        characters_log.info_f("Skipped {} (larger than a character record).", filename);
+        continue;
+      }
       CvtCharacterToPc(&ch);
       characters.emplace_back(std::make_pair(filename, ch.level));
     } catch (phosg::io_error const& error) {


src/realmz_orig/partyselect-helpers.cpp
```diff
@@ -32,7 +32,15 @@ extern "C" void update_character_files_list() {
   for (const auto& filename : mac_list_directory(":Character Files")) {
     try {
       auto f = mac_fopen_unique(std::format(":Character Files:{}", filename), "rb");
+      if (!f) {
+        characters_log.info_f("Skipped {} (could not be opened).", filename);
+        continue;
+      }
       auto ch = phosg::freadx<struct character>(f.get());
+      if (!phosg::fread(f.get(), 1).empty()) {
```
jpetrie, line 40: (1) Why not use `feof()` to test if the underlying file is at the end?

(2) I'm not following how this would "read garbage" that crashes later - `freadx()` will only read up to `sizeof(T)` (in this case, `character`) bytes. Do you mean that there's other code later on that reads a file by reading it in its entirety rather than reading only the first `sizeof(character)` bytes? If that's the case, is it worth addressing those areas as well (in the admittedly unlikely event that a character file is swapped out between here and there)?

chromancer: 1: to hit `feof() `we'd have to attempt read past 872b, in which case` feof()` is redundant to reading one trailing byte of which we already have the return value

2: Yes, that's what I mean saying it triggers on add to party, not selecting the file. When you hit "Add" to add to party on _any_ garbage file that is selected, the null terminators aren't in the right place when it eventually gets read, so you get a buffer overflow when strcat() is unbounded

Under this PR, the path for that is narrow: it must pass all the existing guards (and therefore be listed) by being exactly 872b during select, then either a) still be a garbage or corrupt file that just _happens_ to be exactly the expected size or b) have been swapped for a garbage or corrupt file after selection but before the second read.

to your point that this doesn't completely remove the crash path. We could guard before the first possible buffer overflow `strcat()` of the name on a purely heuristic basis: if we don't see the terminator where it should be, assume the file is garbage, and object to the user.

the natural place to do so I would guess would be at partyselect.c 688, but there's no great existing authentic `warn()` to use; in the interest of not overcomplicating this I'd probably just use 73 and log the actual reason: we think the file selected is garbage or corrupt. 

I would probably draw the line there though: there are very many sites that assume the character struct is correctly sized and correctly terminated throughout, and we could guard them all, but at that point we're hardening against a really improbable file or series of actions by the user.

Make sense to you? If no further thoughts, I'll default to adding the name termination check as above in another commit

jpetrie: Makes sense. Based on that analysis I'm not sure I'd say even the `strcat()` issue is worth addressing once this PR goes in.

jpetrie (approved):

PR #263, opened 2026-06-22
Fix the post-fight treasure screen

afkelsall:
The post-fight treasure screen leaves sparkle and right-edge artifacts behind and replays a stray text blip sound.

Why:
- The item-keep sparkle erases itself with XOR ovals that cancel exactly on the original Mac, but the shim draws antialiased ovals, so stray coloured pixels stay in the cell.
- The sparkle grows past its cell, so at the right edge of the grid the leftovers land outside any tile and are never repainted.
- In the large-window layout `quickinfo` cleared `inbooty` after drawing an item description, so the text blip sound replayed whenever the cursor then left the white item area.
- The artifacts are original drawing code in `src/realmz_orig/` interacting with the shim's antialiased rendering; the `inbooty` clear is an original game logic bug. Both affect macOS and Windows equally.

How:
- After the sparkle, repaint the emptied cell and its neighbours from the item list, widening the cleared region to the sparkle's full extent so the right edge comes clean.
- Reset the hover oval state after the repaint so a stale XOR oval is not drawn back onto the cell.
- Have `quickinfo` save and restore `inbooty` instead of forcing it to `FALSE`.
- Touches only shared game code, so it behaves the same on macOS and Windows.


iSynic:
I went through #248 and #263 carefully. 
I tested current main, this PR, and a narrower artifact-cleanup build in Release. The sparkle cleanup definitely works, and the inbooty preservation makes sense independently.
I also instrumented MouseDownPending(), but it returned false for every pickup I tested and I never saw the flourish get skipped. Looking at the implementation, it's checked only once before entering the blocking 24-frame flourish. That means it can skip the flourish if another click is already queued at that exact moment, but it can't notice a click made while the flourish is playing. The SystemTask() change can drain already-queued events without the additional 10 ms delay, but only after the current handler returns.
So I think the visual/audio fixes are well supported, but I haven't been able to reproduce the rapid-click improvement described here.

afkelsall:
> ...So I think the visual/audio fixes are well supported, but I haven't been able to reproduce the rapid-click improvement described here.

Thank you for the thorough review. The missed clicks are quite irritating and I'll make a new PR to address those in isolation. You're right that logging from the approach in this PR appears to have no impact, and it may very well be the fix in this PR does nothing. But a little niggling feeling in me that after I abandoned #248 and played the game for a bit, I came back and made this PR for a reason, and that reason is still present. I'm still seeing occasional lag and missed clicks on the treasure screen.

I've been able to observe that most clicks animate and finish and are ready for the next click in 100-130ms and my "fast" click rate is 180-250ms. So usually there's no issue. However, from time to time the sparkle animation spikes to 250ms and I *should* see a queued click in that, but do not. It is being flushed and dropped and I believe any logging shows a queue of 0 even though there was _at one point_ something in the queue.

More in #304

src/realmz_orig/booty.c
```diff
@@ -1623,6 +1623,94 @@ void booty(short mode) {
                       PenMode(2);
                       FrameOval(&oldbox);
                       PenMode(0);
+
```
iSynic, line 1626: Sorry I missed this. We need a `CHANGED FROM ORIGINAL IMPLEMENTATION` marker here.

src/realmz_orig/main.c
```diff
@@ -2169,13 +2180,13 @@ void quickinfo(int who, int itemnumber, int itemid, int where) {
...
-    inbooty = FALSE;
+    inbooty = savedinbooty;
   } else {
     getselection(item.itemid);
     inbooty = TRUE; //*** keeps the beep for text from happening
     pict(218, pictrect);
     textbox(tempselection + 2, item.itemid - tempselection + 1, FALSE, TRUE, txtbox);
-    inbooty = FALSE;
+    inbooty = savedinbooty;
   }
```
iSynic, line 2192: Just need to close your changes with the 'end changes' comment down here. 

And if you want, you can update your name on the changed from original implementation on line 2166. I realize that additional guideline came out after this PR just being thorough.

iSynic:
One optional implementation thought while the change markers are being updated... A smaller implementation might be to save the flourish bounds into the existing gbuff2, run the animation, restore that saved region, and then erase only the emptied item cell.
That would avoid manually clearing and redrawing neighboring item cells, duplicating the item-grid rendering logic, and calling loaditem() during cleanup.

But your implementation works, so as far as I'm concerned it's non-blocking and optional for a revisit.

afkelsall:
All good thanks for the review. I'll leave the refactor for a followup dedicated to optimizing this screen further, where it can be tested in isolation. Especially as we're hustling for a beta release. I think this is in a good state to build on and improve further.

iSynic (approved):
Thanks, this addresses the changes I requested. 
I tested the remaining behavior in Release: the sparkle artifacts are cleaned up and inbooty is preserved correctly. My smaller cleanup suggestion is strictly optional and will be looked at in a future PR. Looks good to me.

danapplegate (approved):

PR #264, opened 2026-06-22
Gray out disabled items in Windows popup menus

### LINKED ISSUE #260: "Party members who can use it" menu shows whole party
Doesn't seem to filter on the actual party members who can use the item.

Image

afkelsall:
**What**
- Windows popup menus passed `(item.enabled ? MF_ENABLED : 0)` to `AppendMenu`. `MF_ENABLED` is `0`, so disabled items were never grayed and stayed clickable.
- Now passes `MF_GRAYED` for disabled items.

**Why**
- `DisableItem()` set `enabled = false` and the flag reached the Windows item, but it was dropped at `AppendMenu`.
- Most visible in the "Characters Who Can Use This Item" popup, which showed the whole party as selectable instead of graying out members who can't use the item.
- Distinct from the `SetItemMark`/checkmark work in #236.

Closes #260.

### DIFF
--- a/src/windows/WinMenuController.cpp
+++ b/src/windows/WinMenuController.cpp
@@ -1,14 +1,34 @@
 #include "errhandlingapi.h"
 #include "windef.h"
+#include "wingdi.h"
 #include "winuser.h"
+#include <algorithm>
+#include <cstdint>
+#include <exception>
 #include <memory>
+#include <phosg/Image.hh>
 #include <phosg/Strings.hh>
+#include <vector>
 
 #include "../PortMenu.hpp"
 #include "./WinMenuController.hpp"
 #include <utility>
 
 static phosg::PrefixedLogger wmc_log("[WinMenuController] ");
+static constexpr char kPopupDiamondMark = 19;
+static constexpr size_t kPopupMarkGutterWidth = 12;
+static constexpr int kPopupMarkRadius = 3;
+static constexpr uint32_t kPopupMarkPixel = 0xFF000000;
+static constexpr uint32_t kPopupMixedMarkPixel = 0xFF808080;
+
+enum class PopupMenuMark {
+  None,
+  Checked,
+  Mixed,
+};
+
+// QuickDraw.hpp declares Mac QuickDraw APIs whose names collide with Win32 headers here.
+phosg::ImageRGBA8888N DecodeCIconImage(int16_t iconID);
 
 // Command IDs for the Port menu. They sit in a reserved high range so they never
 // collide with the packed (menu_id, item_id) identifiers the game's own menus use,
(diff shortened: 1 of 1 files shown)


chromancer:
ah I see! I was misremembering. still feature parity but not *marked* but *disabled*. My build was totally broken while i'm in a fugue so I couldn't check.

Screenshot 2026-06-22 at 11 51 33 AM

You may want to consider this is what it looks like mac-side at the moment

afkelsall:
Aww damn the mac looks so much better!
Thanks for the screenshot. I don't know if I have it in me to try to replicate the mac menus on windows - I'll put it on my todo. Maybe I'll raise an issue.

iSynic:
https://github.com/iSynic/realmz/tree/fix-windows-popup-menu-icons

Just as an option, I took another pass at the Windows popup menu behavior here.

This keeps the disabled popup behavior from #264, but also renders the existing Realmz popup menu icons and character indicator marker. It still uses native Windows popup menus rather than owner-drawing the menu.

image

image

The current version keeps `AppendMenu()` for the normal row text/enabled state, then attaches icon bitmaps with `SetMenuItemInfo(... MIIM_BITMAP ...)`. For the current/source character marker, it composes a small fixed-width marker gutter into the icon bitmap itself. That avoids the row offset problem I ran into when trying to use native checkmark bitmaps separately.

So the “Characters Who Can Use This Item” popup now has:

- header/item icon
- character portraits
- disabled/grayed unusable characters
- a small marker for the source/current character
- normal native Windows menu behavior otherwise

I’m not sure if you want to fold this into #264 or keep #264 as the narrower disabled-item fix and handle icons/indicator separately, but this is the cleanest way I could find to reach visual parity.

afkelsall:
Thanks @iSynic - given your menu items are just up in a branch currently, I've merged your branch into this one and will test this locally shortly.

afkelsall:
Tested this branch. The icons in the menu and greying out is working very nicely. I like as well that the icons get greyed out as well when a party member can't use them. Thanks @iSynic for the icon work!
image

chromancer:
@iSynic  would you mind doing build and test portion of review on this?

iSynic:
Thanks for pulling this in. I took one more pass through the current implementation, and I think your decision to keep the marker separate from the Realmz icon is actually more complete than the version on my branch. Realmz also uses `SetItemMark()` in popup rows that don't have an icon, so tying the diamond directly to the icon bitmap wouldn't cover every case.

The one thing I still notice is the pale square beside each disabled row in the latest screenshot. I thought that came from using `kOpaqueWhitePixel` for the disabled unchecked `MIIM_CHECKMARKS` bitmap but I was wrong. Digging into that now.

I can just barely make it out in your screenshot, but on mine it's obvious:

image

I wonder if the cleaner option is to keep `MFS_CHECKED`, but let Windows draw its native checkmark instead of creating the four custom checked/unchecked bitmaps. That would still support marked rows without icons, while removing the blank bitmap issue and quite a bit of the marker/diamond-specific code. I like the Diamond, but I just don't know that we need it if it's a hassle.

I also checked the modern macOS path. It already handles the icons and disabled state, but it doesn't currently apply `item.checked` or `mark_character`, so I would assume the source-character marker is still missing there. But that's should be a different PR altogether by somebody who can actually test it.

chromancer:
w/r/t the nature of the marks, I intend to implement this Mac side with checks and dashes per native menu practice, when I eventually get in front of my computer again

Those were native Mac glyphs and fairly common, but they don't quite look right in a current Mac native menu

Don't let that affect your thinking here necessarily, just letting you know mine. No opinion what is right for Win

iSynic:
I've been trying to implement native Windows checkmark handling, but the portrait bitmap makes it tough.

`MFS_CHECKED` works correctly on a normal text-only menu item. As soon as the same item also has an `hbmpItem` portrait, Windows stops displaying the native checkmark.

Realmz is passing the checked state correctly. The problem is that Win32 uses the same left-side menu area for both the item bitmap and the native check indicator. So on rows with a portrait, only the portrait is drawn.

Back to the drawing board for a bit.

iSynic:
I think I've got a working version without artifacts and have [replaced the branch I linked above](https://github.com/iSynic/realmz/tree/fix-windows-popup-menu-icons) with my latest version if you want to take a look or try it out. 

I tried native checks, separate MIIM_CHECKMARKS bitmaps, and owner-drawn rows. Those either lost the marker, shifted only the checked row, caused the square artifacts, or required substantially more custom menu rendering.

This version:

- attaches the existing Realmz icons and portraits
- keeps disabled rows and portraits grayed normally
- reserves a small, consistent gutter inside icon-bearing rows
- draws the black diamond or mixed-state dash in that gutter
- uses the native Windows checkmark/dash for marked rows without icons

I tested the character-eligibility popup, equipped-item checks, and a marked condition row without an icon. All three now display and line up.

image

image

afkelsall:
Wow yep you are completely right my screenshot still had the boxes! I've tested your branch on my system and can confirm the boxes are gone and the other stuff remains. Nice one! This really is all your work now, do you want to open a PR and reference this one and I'll close this and reference yours? I'm happy to update this PR to have your changes in it if you want to, but happy also to attribute this work to you properly through your own PR :) 

image

iSynic (approved):
This appears to work for @afkelsall and myself. So, approved.

danapplegate (approved):
Nice work @afkelsall and @iSynic! Looks fantastic.

PR #266, opened 2026-06-22
Fix crash plotting cicn icons that lack a monochrome bitmap

afkelsall:
Fixes an access-violation crash that happens mid-combat when the spell-info panel draws certain target-type icons.

Why:
- Lives in the SDL compatibility shim (`src/QuickDraw.cpp`); it is the same defect on macOS and Windows since both share this code path.
- A `cicn` resource's 1-bit monochrome bitmap is optional and is absent for most of the game's icons. When it is missing, `decode_cicn` returns an empty bitmap.
- `GetCIcon` still allocated a zero-size `bitmapData` handle and left `iconBMap.bounds` at the color image's size, so `PlotCIconBitmap` blitted a color-sized region out of the empty buffer and read past it into unmapped memory.

How:
- Only allocate `bitmapData` when the decoded monochrome bitmap has data; leave it `nullptr` otherwise.
- Make `PlotCIconBitmap` a no-op when `bitmapData` is `nullptr`.
- Keep `iconBMap.bounds` at the color image size, which the game reads for icon layout.

Repro: in combat, open the spell-info display for a group or area spell (all enemies, all friendly, self) whose target-type icon has no monochrome bitmap.

### DIFF
--- a/src/QuickDraw.cpp
+++ b/src/QuickDraw.cpp
@@ -818,13 +818,20 @@ CIconHandle GetCIcon(uint16_t iconID) {
 
   CIconHandle h = NewHandleTyped<CIcon>();
   (*h)->iconData = NewHandleWithData(decoded_cicn.image.get_data(), decoded_cicn.image.get_data_size());
-  (*h)->bitmapData = NewHandleWithData(decoded_cicn.bitmap.get_data(), decoded_cicn.bitmap.get_data_size());
+  // The monochrome bitmap is optional and absent for most cicn resources; when
+  // it is missing decode_cicn returns an empty bitmap. Leave bitmapData null in
+  // that case so PlotCIconBitmap knows there is nothing to draw.
+  (*h)->bitmapData = decoded_cicn.bitmap.get_data_size()
+      ? NewHandleWithData(decoded_cicn.bitmap.get_data(), decoded_cicn.bitmap.get_data_size())
+      : nullptr;
   (*h)->iconPMap.bounds = Rect{
       0,
       0,
       static_cast<int16_t>(decoded_cicn.image.get_height()),
       static_cast<int16_t>(decoded_cicn.image.get_width())};
   (*h)->iconPMap.pixelSize = 32;
+  // iconBMap.bounds tracks the color image size because the game reads it for
+  // layout; it stays valid even when there is no monochrome bitmap.
   (*h)->iconBMap.bounds = (*h)->iconPMap.bounds;
   return h;
 }
@@ -858,11 +865,15 @@ OSErr PlotCIcon(const Rect* r, CIconHandle icon) {
 }
 
 OSErr PlotCIconBitmap(const Rect* r, CIconHandle icon) {
+  auto& port = current_port();
+  port.log.debug_f("PlotCIconBitmap({{x0={}, y0={}, x1={}, y1={}}}, {:p})", r->left, r->top, r->right, r->bottom, static_cast<void*>(icon));
+  // The cicn may have no monochrome bitmap, in which case there is nothing to draw.
+  if ((*icon)->bitmapData == nullptr) {
+    return noErr;
+  }
   auto bounds = (*icon)->iconBMap.bounds;
   int w = bounds.right - bounds.left;
   int h = bounds.bottom - bounds.top;
-  auto& port = current_port();
-  port.log.debug_f("PlotCIconBitmap({{x0={}, y0={}, x1={}, y1={}}}, {:p})", r->left, r->top, r->right, r->bottom, static_cast<void*>(icon));
   port.draw_ga11_data(*((*icon)->bitmapData), w, h, *r);
   WindowManager::instance().recomposite_from_window(port);
   return noErr;


afkelsall:
The first commit had a display error on a few things I noticed quickly. The monochrome bitmap in a cicn is optional and is absent for most of the game's icons, in which case decode_cicn returns an empty 0x0 bitmap.

icon-fix-display-issues

danapplegate:
I'm having a bit of a hard time following the PR's (Claude's?) logic.

- I think the "Original vs PC port" section is an hallucination, since both Mac and Windows builds of the port follow the same code path now. So the crash should be present on both.
- It looks like tried to change the logic in `GetCIcon`, but reverted back to the current implementation which always sets `iconBMap.bounds` equal to `iconPMap.bounds`. This means that the `w <= 0 || h <= 0` part of the check added to `PlotCIconBitmap` is useless, because `w` and `h` will never be 0.
- If that's the case, then the real meat of this fix is just the part which checks that the bitmap handle has no data, `GetHandleSize((*icon)->bitmapData) == 0`.
- Digging in a bit further, it looks like the MemoryManager would end up calling `malloc(0)` in this case, which [returns a `nullptr` as the data handle](https://github.com/Realmz-Castle/realmz/blob/3516240934ed3e9c38e6454782ec320b3577f501/src/MemoryManager.cpp#L106-L110) (and also sets `this->last_error = memFullErr`, which seems incorrect). So, I think it's slightly cleaner here to check whether `(*icon)->bitmapData == nullptr`, rather than indirectly checking so by looking at its size.

@fuzziqersoftware Does that sound right to you??

danapplegate:
@afkelsall Also, do you have a particular spell or save file that you use to reliably reproduce this?

afkelsall:
This happened to me in CoB spider tower I think in one of the battles inside but after the main encounter. I don't think it was against a particular magic caster from memory but may have been something I'd summoned. It happened about 3 hours into a session but I wasn't running on debug. 

I found this issue from looking at the windows crash dump. There's only been one crash I've been able to reliably produce which is when you go behind the magistrate through the secret wall, loot her stash and the trigger the encounter on the way back out. That was a different problem though.

afkelsall:
I can dig up the dump I think still and also share the transcript if it helps.

iSynic:
> I'm having a bit of a hard time following the PR's (Claude's?) logic.
> 
>     * I think the "Original vs PC port" section is an hallucination, since both Mac and Windows builds of the port follow the same code path now. So the crash should be present on both.
> 
>     * It looks like tried to change the logic in `GetCIcon`, but reverted back to the current implementation which always sets `iconBMap.bounds` equal to `iconPMap.bounds`. This means that the `w <= 0 || h <= 0` part of the check added to `PlotCIconBitmap` is useless, because `w` and `h` will never be 0.
> 
>     * If that's the case, then the real meat of this fix is just the part which checks that the bitmap handle has no data, `GetHandleSize((*icon)->bitmapData) == 0`.
> 
>     * Digging in a bit further, it looks like the MemoryManager would end up calling `malloc(0)` in this case, which [returns a `nullptr` as the data handle](https://github.com/Realmz-Castle/realmz/blob/3516240934ed3e9c38e6454782ec320b3577f501/src/MemoryManager.cpp#L106-L110) (and also sets `this->last_error = memFullErr`, which seems incorrect). So, I think it's slightly cleaner here to check whether `(*icon)->bitmapData == nullptr`, rather than indirectly checking so by looking at its size.
> 
> 
> @fuzziqersoftware Does that sound right to you??

I think this is the right read.

I tested this locally with a small fixture using a real bundled `cicn` that has color image data but no monochrome bitmap data and confirmed it's an issue on Main but that your proposed fixes work.

On current main, `GetCIcon()` still allocates `bitmapData` for the empty monochrome bitmap. 

On the smaller fix, `GetCIcon()` leaves `bitmapData == nullptr`, keeps `iconBMap.bounds` tied to the color image bounds, and `PlotCIconBitmap()` returns `noErr` when there is no monochrome bitmap to draw.

So I think the core fix can be reduced to:

- only allocate `bitmapData` when the decoded monochrome bitmap has data
- keep `iconBMap.bounds` from the color image, since Realmz appears to use that for layout
- make `PlotCIconBitmap()` no-op when `bitmapData == nullptr`

That keeps this focused on optional `cicn` bitmap data without changing MemoryManager behavior globally.

danapplegate:
> So I think the core fix can be reduced to:
>
> - only allocate bitmapData when the decoded monochrome bitmap has data
> - keep iconBMap.bounds from the color image, since Realmz appears to use that for layout
> - make PlotCIconBitmap() no-op when bitmapData == nullptr
>
> That keeps this focused on optional cicn bitmap data without changing MemoryManager behavior globally.

Agreed, that sounds like the right approach! I think that also means much of the added comments can be removed or simplified.

fuzziqersoftware:
Sorry for my delay in responding - yeah, I agree with Dan here. I think we should explicitly check in GetCIcon whether the bitmap image is empty, and if so, set bitmapData to null. It's odd that Realmz uses iconBMap even if there's no actual bitmap, but we can still populate it. I suppose this behavior is technically valid because Classic Mac OS checks flags_row_bytes instead of the bitmap rect to determine whether there's a bitmap or not (and this is what resource_dasm does as well) so the bitmap rect can still be valid in that case.

danapplegate (approved):
Thanks for finding and fixing this crash! Fix looks good to me.

PR #270, opened 2026-06-24
Fix fumbled item drops

### LINKED ISSUE #254: Fumbled item drops can scan past inventory and overrun the fumble queue
There are three player fumble paths that scan inventory with `fumloop <= numitems`:

- weapon fumble during attacks
- encounter opcode `122` cause fumble
- spell special `67` fumble

Those paths also append to `fumque[20]` without checking `fumtotal`, and they read the fumbled item id again after calling `removeitem()`.

That looks like it can crash or corrupt data if the loop runs past valid inventory slots, if the fumble queue fills, or if the item reference changes after unequipping.

Test fix branch here:
https://github.com/iSynic/realmz/tree/fix-fumbled-item-drop

This is based on a broader fix I had in another fork, but trimmed down to just the fumble paths to keep a potential PR narrow. I did catch the weapon fumble during attacks crash live on a debug build, which led to my search.

Opening this for comments before submitting.

iSynic:
Fixes #254.

Some player fumble paths scan inventory with `fumloop <= numitems`, which can look one slot past the active inventory. They also append to `fumque[20]` without checking whether the fumble queue is already full.

This tightens the three player item fumble paths:

- weapon fumble during attacks
- encounter opcode 122
- spell special 67 (may not be used by scenarios but uncertain)

The scan now stays within the active inventory and the 30-slot inventory limit, and the fumble queue is checked before adding another item. The matched item id is saved before the unequip/drop sequence so the queued and dropped item stay explicit.

### DIFF
--- a/src/realmz_orig/attack.c
+++ b/src/realmz_orig/attack.c
@@ -204,13 +204,20 @@ moveon:
 
     if ((templong > 50 && templong < 60) && (c[chare].armor[2])) /**** fumble weapon ****/
     {
-      for (fumloop = 0; fumloop <= c[chare].numitems; fumloop++) {
+      short fumbleditem;
+      /* *** CHANGED FROM ORIGINAL IMPLEMENTATION ***
+       * Guard fumbled item drops against inventory and fumble queue bounds.
+       */
+      for (fumloop = 0; (fumloop < c[chare].numitems) && (fumloop < 30); fumloop++) {
         if (c[chare].items[fumloop].id == c[chare].armor[2]) {
+          if (fumtotal >= 20)
+            goto donefumble;
+          fumbleditem = c[chare].items[fumloop].id;
           if (removeitem(chare, fumloop, FALSE, FALSE)) {
             sound(-10121);
             sound(-10123);
-            fumque[fumtotal++] = c[chare].items[fumloop].id;
-            dropitem(chare, c[chare].items[fumloop].id, fumloop, TRUE, FALSE);
+            fumque[fumtotal++] = fumbleditem;
+            dropitem(chare, fumbleditem, fumloop, TRUE, FALSE);
             c[chare].armor[2] = c[chare].weaponnum = 0;
             warn(114);
             success = FALSE;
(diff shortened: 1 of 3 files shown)


src/realmz_orig/attack.c
```diff
@@ -13,7 +13,7 @@ short attack(short chare, short mon) {
   struct character character;
   struct monster monst;
   short att, specialdam;
-  short t;
+  short t, fumbleditem;
```
jpetrie: This should be marked with a "changed from original" comment given how far it is from the actual change that uses it.

Alternatively, I think `fumbleditem` should be able to get moved to the top of the `/**** fumble weapon ****/` block where it is actually used safely, and that would still comply with C90 rules on declarations before other statements within a block.

(We also appear to have some extension enabled somewhere that just allows arbitrary declarations anywhere, like in C++, but I'm not sure if that's intentional and if it would interfere with cross-compilation.)

This applies to the other two introductions of `fumbleditem` as well.

src/realmz_orig/attack.c
```diff
@@ -204,13 +204,19 @@ short attack(short chare, short mon) {
 
     if ((templong > 50 && templong < 60) && (c[chare].armor[2])) /**** fumble weapon ****/
     {
-      for (fumloop = 0; fumloop <= c[chare].numitems; fumloop++) {
+      /* *** CHANGED FROM ORIGINAL IMPLEMENTATION ***
+       * Guard fumbled item drops against inventory and fumble queue bounds.
+       */
+      for (fumloop = 0; (fumloop < c[chare].numitems) && (fumloop < 30); fumloop++) {
```
jpetrie, line 211: At some point (not needed for this PR), we should introduce constants for the 30 and 20 limits.

jpetrie (changes requested):

src/realmz_orig/attack.c
iSynic: Thanks for the review and feedback. This should be addressed.

jpetrie (approved):

PR #272, opened 2026-06-25
Automatically copy resource file dependencies.

### LINKED ISSUE #168: Launching the non-packaged executable crashes ("current resource file is out of bounds")

jpetrie:
This PR fixes #168 by ensuring that resource files Realmz needs are copied to the build output directory automatically, removing the need for manual copies or symlinks on the part of the developer.


CMakeLists.txt
```diff
@@ -260,8 +260,8 @@ set(RESOURCE_FILES
     bundle/AppIcon.icns
     "resources/Black Chancery.ttf"
     "resources/Black Chancery Bold.ttf"
-    "resources/Inter/InterVariable.ttf"
     "resources/ChicagoFLF.ttf"
+    "vendored/Inter/InterVariable.ttf"
```
jpetrie, line 264: Moving this file makes the CMake copy commands much simpler. It is a third-party project, so `vendored/` seemed like the best place to do so while still making the associated license file obvious.

CMakeLists.txt
```diff
@@ -336,6 +336,13 @@ if(NOT WIN32 AND NOT CMAKE_BUILD_TYPE STREQUAL "Release")
     target_link_options(Realmz PRIVATE -fsanitize=address)
 endif()
 
+# Ensure the files that Realmz needs exist next to the executable.
+add_custom_command(TARGET Realmz POST_BUILD
+    COMMAND ${CMAKE_COMMAND} -E copy_directory "${CMAKE_CURRENT_SOURCE_DIR}/resources" "$<TARGET_FILE_DIR:Realmz>"
+    COMMAND ${CMAKE_COMMAND} -E copy_directory "${CMAKE_CURRENT_SOURCE_DIR}/base/Realmz" "$<TARGET_FILE_DIR:Realmz>"
```
danapplegate, line 342: It looks like `copy_directory` allows you to specify multiple source directories before specifying the target, so I think these two lines can be combined.

danapplegate (commented):
Nice work!

I agree with your original assessment from #168, that setting up a bundle build again would be the cleanest, because it would allow us to make use of the same configured install components for both the distributable and the bundle. So it's a bit awkward to have to manually copy them here in this post-build command, in addition to specifying that they need to be installed during packaging.

With that said, I doubt we'll have to update these regularly (ever again?), so I'm fine with the duplication.

danapplegate (approved):

Issue #274, opened 2026-06-25
Discuss impact of fixing item + to hit rating in combat


afkelsall:
As pointed out in discord by @iSynic (I believe), it's not just the weapon + to hit rating that's missing from combat() calcs, items are also missing.

The combination of both these changes potentially alters the game. This discussion is to establish the current behavior across levels and items against various difficulty monsters. Then assess the impact of fixing both weapon and item to hit in combat.

For me, take a Helm of Might +2. If you take away the hit rating, it's a standard +1 helm with +1 damage. Maybe that's enough for it to be "mighty", but the tooltip has +5% to hit rating that's currently not added in combat.

I'm personally in two minds about this change. I'm interested to test it. When an item says it adds +hit but doesn't, that doesn't feel right.

Here are the current hit tables before we make any changes:
Columns for this comparison are:
- `Bare (+5)` = No magical items - +5 to hit gained per level
- `Items (+5)` = Helm of Might (+5 to hit) + Gauntlets of Damage (+5 to hit) + Combat Boots (+5 to hit) + Regular +2 weapon (+10 to hit)
- `Items+Str (+5)` = Above + War Axe of Strength (strength = +15 to hit)
- `Items+Str+Ring (+5)` = Above + Ring of Combat (+30 to hit)

Within each cell, the first number is against a weak monster (AC25), middle is medium (AC50 - around Blue Dragon/Dinosaur/Terror beast) and right is tough (AC75 - hell dragon in CoB)
```
┌─────┬──────────────┬───────────────┬───────────────────┬────────────────────────┐
│ Lvl │ Bare (+5)    │ Items (+5)    │ Items+Str (+5)    │ Items+Str+Ring (+5)    │
├─────┼──────────────┼───────────────┼───────────────────┼────────────────────────┤
│   1 │ 40 / 15 / 5  │  40 / 15 / 5  │    55 / 30 / 5    │      55 / 30 / 5       │
├─────┼──────────────┼───────────────┼───────────────────┼────────────────────────┤
│   5 │ 60 / 35 / 10 │ 60 / 35 / 10  │   75 / 50 / 25    │      75 / 50 / 25      │
├─────┼──────────────┼───────────────┼───────────────────┼────────────────────────┤
│  10 │ 85 / 60 / 35 │ 85 / 60 / 35  │   100 / 75 / 50   │     100 / 75 / 50      │
├─────┼──────────────┼───────────────┼───────────────────┼────────────────────────┤
[...]

Issue #269, opened 2026-06-23
Weapon magic plus to melee to-hit is not used in attack calculation for player characters.


afkelsall:
Player melee attacks ignore a weapon's magic plus when rolling to hit, so a +N weapon does not improve a character's hit chance.

Why:
- `attack()` adds `item.damage` (magic plus) to melee damage but never to the to-hit roll.
- `attack2()` (monster) and `resolvespell()` (missile) both add 5 per point, and the character sheet attack bonus already shows it (`damage * 5`).
- Player melee is the only path that drops it, so the sheet and combat disagree.

How:
- Add `att += 5 * item.damage` in `attack()`, after the damage bonus and before the penetration check, matching attack2/resolvespell.
- Penetration weapons (sp1 == 121) stack the base term with their existing bonus, giving the intended doubling.

Verification:
- Logging from equip and attack code.
- Equip is to see the items stats as it's equipped.
- The top attack of each weapon is the original behavior.
- The bottom attack is with the proposed fix.
- Same character, same monster, same weapon, same battle, same round even - two attacks after each other with fix off then on. Jitter is to do with luck rating mainly.
- Compare each attacks `att_from_weapon` and `att_final`.

```
Penetration tip sword +2
[EQUIP  ] char=1 item_id=74 plus(item.damage)=2 sp1=121 sp2=0 item.st=0 => c.tohit=20 c.damage=2 c.magst=1 strong=0
[ATTACK ] chare=1 mon=16 weap_id=74 plus=2 sp1=121 fix=0 | char.tohit=23 char.damage=2 strong=0 behind=0 | att_from_weapon=10 att_final=72
[ATTACK ] chare=1 mon=16 weap_id=74 plus=2 sp1=121 fix=1 | char.tohit=23 char.damage=2 strong=0 behind=0 | att_from_weapon=20 att_final=85

war axe of strength +2
[EQUIP  ] char=1 item_id=86 plus(item.damage)=2 sp1=41 sp2=-1 item.st=0 => c.tohit=23 c.damage=2 c.magst=1 strong=0
[ATTACK ] chare=1 mon=17 weap_id=86 plus=2 sp1=41 fix=0 | char.tohit=23 char.damage=2 strong=-1 behind=0 | att_from_weapon=0 att_final=74
[...]

PR #268, opened 2026-06-23
Apply weapon magic plus to player melee to-hit

afkelsall:
Fixing bug #269.

Player melee attacks ignore a weapon's magic plus when rolling to hit, so a +N weapon does not improve a character's hit chance.

Why:
- `attack()` adds `item.damage` (magic plus) to melee damage but never to the to-hit roll.
- `attack2()` (monster) and `resolvespell()` (missile) both add 5 per point, and the character sheet attack bonus already shows it (`damage * 5`).
- Player melee is the only path that drops it, so the sheet and combat disagree.

How:
- Add `att += 5 * item.damage` in `attack()`, after the damage bonus and before the penetration check, matching attack2/resolvespell.
- Penetration weapons (sp1 == 121) stack the base term with their existing bonus, giving the intended doubling.

Verification:
- Logging from equip and attack code.
- Equip is to see the items stats as it's equipped.
- The top attack of each weapon is the fix toggled off.
- The bottom attack is the fix toggled on.
- Same character, same monster, same weapon. Jitter is to do with luck rating mainly.
- Compare each attacks `att_from_weapon` and `att_final`.

```
Penetration tip sword +2
[EQUIP  ] char=1 item_id=74 plus(item.damage)=2 sp1=121 sp2=0 item.st=0 => c.tohit=20 c.damage=2 c.magst=1 strong=0
[ATTACK ] chare=1 mon=16 weap_id=74 plus=2 sp1=121 fix=0 | char.tohit=23 char.damage=2 strong=0 behind=0 | att_from_weapon=10 att_final=72
[ATTACK ] chare=1 mon=16 weap_id=74 plus=2 sp1=121 fix=1 | char.tohit=23 char.damage=2 strong=0 behind=0 | att_from_weapon=20 att_final=85

war axe of strength +2
[EQUIP  ] char=1 item_id=86 plus(item.damage)=2 sp1=41 sp2=-1 item.st=0 => c.tohit=23 c.damage=2 c.magst=1 strong=0
[ATTACK ] chare=1 mon=17 weap_id=86 plus=2 sp1=41 fix=0 | char.tohit=23 char.damage=2 strong=-1 behind=0 | att_from_weapon=0 att_final=74
[ATTACK ] chare=1 mon=17 weap_id=86 plus=2 sp1=41 fix=1 | char.tohit=23 char.damage=2 strong=-1 behind=0 | att_from_weapon=10 att_final=90

 morningstar +2 - straight +2 weapon
[...]

PR #275, opened 2026-06-25
Apply worn item magic plus to player melee to-hit

afkelsall:
A worn item or armor magic plus, such as a Helm of Might +2, advertises a to-hit bonus on its label but never affected the player melee to-hit roll.

Why:
- attack() only added the weapon's magic plus (5 per point) to the to-hit roll (previous fix in #268)
- Every worn item plus is summed into character.damage by wear, and that total already raised the damage dealt and the displayed attack bonus (character.damage * 5 on the character sheet), but it never reached the to-hit roll.
- So a +N helm or armor showed a to-hit bonus on its label and in the attack bonus, yet contributed nothing when rolling to hit; only the weapon's plus and strength items did.

How:
- Base the to-hit term on character.damage instead of the weapon's item.damage, in the shared moveon path after the base tohit term.
- character.damage already includes the weapon's plus, so this one term covers the weapon and every other worn item, matching the attack bonus on the character sheet and the damage path.
- The penetration weapon (sp1 == 121) still adds an extra 5 * item.damage for its intended doubling.
- Since character.damage applies whether or not a weapon is equipped, an unarmed attacker now also gets the item plus on to-hit, consistent with the damage path and the displayed bonus.
- Change is in the shared original game logic, so it stays macOS and Windows compatible.

Follows on from the weapon-only fix in #268 (issue #269). Provides the item half discussed in #274.

### DIFF
--- a/src/realmz_orig/attack.c
+++ b/src/realmz_orig/attack.c
@@ -69,14 +69,6 @@ short attack(short chare, short mon) {
       }
     }
     damage += item.damage; /******* magic plus ****/
-    /* *** CHANGED FROM ORIGINAL IMPLEMENTATION ***
-     * The weapon magic plus (item.damage) was added to player melee damage but
-     * never to the to-hit roll. The monster path (attack2) and missile path
-     * (resolvespell) both add 5 per point, and the character sheet attack bonus
-     * already includes it (damage * 5), so player melee was the only path that
-     * ignored it. Add the same term here. */
-    att += 5 * item.damage; /******* magic plus to hit; matches attack2 and resolvespell ****/
-    /* *** END CHANGES *** */
     if (item.sp1 == 121)
       att += 5 * item.damage; /******* double to hit weapon ****/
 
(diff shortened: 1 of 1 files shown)


afkelsall:
Items and weapons do not separate their +damage from their +to hit rating. The tool tip for damage and to hit in game is rendered from the same `item.damage` field. This almost always corresponds to the shown magic + rating of the item itself. An items + rating corresponds to an equivalent + damage and + to hit - _**except**_ Helm of Might +2 which is actually a +1 item. In fact, perhaps its label is incorrect when comparing to everything else.

Edit: Oh I note "Combat boots +2" is also labeled in the same way. 

image

image

src/realmz_orig/attack.c
```diff
@@ -102,6 +94,19 @@ short attack(short chare, short mon) {
...
+   * that total already raises the damage dealt (damage += character.damage below)
+   * and the displayed attack bonus (character.damage * 5 in updatecharinfo), but it
+   * never reached the to-hit roll. The weapon path (attack2) and missile path
+   * (resolvespell) add 5 per point for the weapon, and the character sheet adds 5
+   * per point for every worn item, so player melee to-hit was the only place an
+   * item plus was ignored. Add 5 per point of character.damage here so a +N helm or
+   * armor contributes to to-hit, not just a +N weapon. character.damage already
+   * includes the weapon's plus, so this covers it too; the penetration weapon still
+   * adds an extra 5 * item.damage above for its intended doubling. */
+  att += 5 * character.damage;
```
afkelsall, line 108: I've been testing and checking this. It's double counting strength stat based to hit in this form. Strength adds to both `character.tohit` and `character.damage`.

jpetrie: Just so I'm clear, is this comment indicative of something you're still trying to work out (e.g., this PR should be a draft) or would you like me to review it as-is?

afkelsall: Yeah sorry I think this should have been a draft, and later removed. I think given all the discussion here, this has been tricky to understand and get right. [This PR](https://github.com/Realmz-Castle/realmz/pull/290) from @chromancer is I think the culmination of all our learning's now and is looking good IMO.

Closing this PR to focus on that one alone now.

afkelsall:
https://github.com/Realmz-Castle/realmz/pull/290 from @chromancer is the culmination of all our learning's now and is looking good IMO.

Closing this PR to focus on that one alone now.

PR #277, opened 2026-06-28
Fix combat crash from out-of-range body id in centerfield

afkelsall:
Starting a combat encounter can crash the game with an access violation partway through the turn.

Why:
- `centerfield` scans 16 columns by 14 rows of `field` starting at `fieldx`, `fieldy`, one more of each than the 15 by 13 the look window shows, so the last column and row are drawn outside `lookrect` and never reach the screen.
- `field` is `short[90][90]`, but the clamp lets `fieldx` reach 75, so the column loop runs to `t == 90` and `field[90][tt]` reads past the end of the array into the globals declared after it.
- Whatever sits there becomes `tempicon`, and any value from 0 to 999 is taken for a body id, since an in-range cell only ever holds terrain (1000 and up) or a real body id.
- That stray id makes `bq[tempicon] = TRUE` write past the end of `bq`, a `char[maxloop]` array on the stack, and overwrite a saved register. The corruption is silent until later in the same turn, when `getup` dereferences the now-bad pointer and the game faults.
- Separately, the `fieldy` clamp of 76 leaves field row 89 unreachable. The bottom visible row is `fieldy + 12`, while the `fieldx` clamp of 75 already allows column 89. `centerpict` clamps at 75 and 77 for the same window.
- This is in the original game logic (`src/realmz_orig/`), not the shim, and not a Mac-versus-Windows difference. The same out-of-bounds read and write exist on every target, and it faults on Windows because of where that toolchain puts the stray byte.

How:
- Bound both scan loops to the field (`t < 90`, `tt < 90`), so the off-screen column and row can no longer read past the array. Nothing visible is lost, since that column and row fall outside `lookrect`.
- Clamp `fieldy` at 77 rather than 76, matching `centerpict`, so the last row of the field can be scrolled into view.
- Keep a bound on the write to `bq`. It is no longer reachable from the field read, but `bq` is a stack array and a bad id from any other source would smash the stack there.
- The change is in shared original code, so it applies to both macOS and Windows.

### DIFF
--- a/src/realmz_orig/centerfield.c
+++ b/src/realmz_orig/centerfield.c
@@ -82,8 +82,19 @@ void centerfield(short x, short y) {
         CopyBits(src, dst, &itemRect, &icon, 0, NIL);
 
       } else if (tempicon > -1) {
-        bodyground(tempicon, 1);
-        bq[tempicon] = TRUE;
+        /* *** CHANGED FROM ORIGINAL IMPLEMENTATION ***
+         * Bound the body id before using it to index bq and the body arrays.
+         * bq is char[maxloop] and only body ids 0..maxloop-1 are valid (party
+         * 0..8, monsters 10..10+maxmon-1). A stray field cell holding maxloop
+         * or more made bq[tempicon] = TRUE write past the end of bq on the
+         * stack, corrupting a saved register and crashing later in the same
+         * combat turn. The redraw loop below already bounds its index with
+         * t < maxloop; match it here so an out-of-range id is skipped. */
+        if (tempicon < maxloop) {
+          bodyground(tempicon, 1);
+          bq[tempicon] = TRUE;
+        }
+        /* *** END CHANGES *** */
       }
     }
   }
--- a/src/realmz_orig/centerfield.c
+++ b/src/realmz_orig/centerfield.c
@@ -32,16 +32,10 @@ void centerfield(short x, short y) {
   if (newx > 75)
     x -= (newx - 75);
   /* *** CHANGED FROM ORIGINAL IMPLEMENTATION ***
-   * Clamp fieldy at 77 rather than 76 so the last row of the field can be
-   * scrolled into view. The look window shows 15 columns by 13 rows, so the
-   * bottom visible row is fieldy + 12; stopping fieldy at 76 left row 89
-   * unreachable while the fieldx clamp of 75 already allowed column 89.
-   * centerpict clamps at 75 and 77 for the same window, which is the pattern
-   * followed here. The redraw loop below bounds its indices to the field, so
-   * the extra row it scans past the window stays in range. */
+   * Clamp fieldy at 77, not 76, so the last field row can scroll into view,
+   * matching the fieldx clamp of 75 and centerpict's 75/77. */
   if (newy > 77)
     y -= (newy - 77);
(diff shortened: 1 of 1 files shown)


afkelsall:
The first commit for this fix has not addressed the problem of the out of bounds index bring written, just this access site. Need to figure out why it was written out of bounds. This led to another crash.

src/realmz_orig/centerfield.c
chromancer: try this

`for (t = fieldx; (t < fieldx + 16) && (t < 90); t++)`

chromancer: per discord convo

chromancer: this solution raises the question: if 0-89 inclusive are visible on `x` why shouldn't `89` be visible on `y`? I think it's another oversight. I can't see any reason why this shouldn't be approximately symmetric with e.g. centerpict.c starting at ~46… strong evidence that's the original intent, so:

```
  if (newx > 75)
    x -= (newx - 75);
  if (newy > 77)
    y -= (newy - 77);
```

then…

`for (tt = fieldy; tt < fieldy + 14 && tt < 90; tt++) // Myriad (+11)` also!

am I wrong?

afkelsall: You are right on both counts, and the first one turned out to be the root cause rather than just a safer access.

The scan loops cover 16 columns by 14 rows starting at `fieldx`, `fieldy`, one more of each than the 15 by 13 the look window shows, so the last column and row are drawn outside `lookrect` and never reach the screen. `field` is `short[90][90]` and the clamp lets `fieldx` reach 75, so the column loop runs to `t == 90` and `field[90][tt]` reads past the end of the array into the globals declared after it. Whatever sits there lands in `tempicon`, and anything from 0 to 999 is then taken for a body id: an in-range cell only ever holds terrain (`combatmap.c` adds 1000 to every cell) or a real body id, never a value in between. So that read is where the stray id came from that overran `bq`. Bounding the loop stops it at the source, and it costs nothing on screen, since that column is outside `lookrect` anyway.

On the `y` side you are right too. The bottom visible row is `fieldy + 12`, so clamping `fieldy` at 76 leaves row 89 unreachable while the `fieldx` clamp of 75 already allows column 89, and `centerpict` uses 75 and 77 for the same window. One nuance: 76 is not what crashes today, since `76 + 13 = 89` is still in range. It only starts to matter once the clamp goes to 77, which is exactly why the `tt < 90` bound has to come with it.

Pushed both bounds, the `fieldy` clamp at 77, and a kept bound on the write to `bq` in case a bad id ever arrives from somewhere else.

chromancer: Just my thoughts! I have just one final quibble on this one, can we cut down the length of the changed from original comments by like 90%? I think we can cover the bugfix with one comment block plus end changes, or one block and two inlines, whatever

chromancer:
Comment on comments is non-blocking. Let me know if you prefer strongly to merge as is.

chromancer (approved):
Have built. No issues in combat resulting from the fix; no crashes found after fix; static analysis leads us to believe this should return us to the stable and originally intended camera clamp behavior that is not capable of OOB read.

Merge approved, pending decision on potentially cutting down comment LOC added

chromancer:
Merged #277 to main.

PR #279, opened 2026-06-28
Draw dBoxProc dialog frames

### LINKED ISSUE #278: Windows dBoxProc dialogs are missing their modal frame
@chromancer noticed that the current codebase draws certain dialog boxes as borderless, and the old Mac-style outer frames are missing and look cropped when compared side by side. 

Mac Classic example at 6x:
Image

Current codebase ported to Windows:
Image

The port currently treats dBoxProc dialogs as effectively borderless. The dialog contents/resources are there, but the Classic Mac dialog chrome implied by dBoxProc is not being drawn on Windows.

PR is live, but I suppose we could discuss whether this is intended/desired. There are other windows with borders that this doesn't handle, but I figure we'd want to be narrow with our scope as we address these. 

For what it's worth, my installed copy of the commercial Fantasoft/Windows version does have (albeit different) borders, and from what I can tell it's also doing something similar to draw them (just not trying to imitate Mac chrome). I'm just not sure when or why it was changed for the borderless approach. See:

Image

iSynic:
Fixes #278.

Windows was rendering `dBoxProc` dialogs with only a one-pixel border, so dialogs such as the monster attack table were missing the Classic Mac frame implied by their resource proc ID.

This draws the frame outside the existing dialog content rect during window recomposition. It uses each dialog's `portRect`, so it is not tied to a particular dialog size and does not resize the dialog or move any DITL items. Dialogs using other window kinds are unchanged.

I also updated the old warning text so it describes unsupported window kinds instead of treating every recognized dialog kind as borderless.

Current Windows build:
Current Windows dialog without dBoxProc frame

Classic Mac:
Classic Mac dBoxProc frame

Updated Windows build:
Updated Windows dBoxProc frame

Edit: I recognize that this is fixing something that's less about Realmz and more about how some of us remember it looking. So if we'd rather just close this one out and work on more important things, I'm good with that.

### DIFF
--- a/src/WindowManager.cpp
+++ b/src/WindowManager.cpp
@@ -56,6 +56,73 @@ inline Rect copy_rect(const ResourceDASM::Rect& src) {
   return Rect{.top = src.y1.load(), .left = src.x1.load(), .bottom = src.y2.load(), .right = src.x2.load()};
 }
 
+static void draw_dbox_frame_line(
+    CCGrafPort& port,
+    const Rect& content_rect,
+    int16_t offset,
+    uint32_t top_left_color,
+    uint32_t bottom_right_color) {
+  int32_t left = content_rect.left - offset;
+  int32_t top = content_rect.top - offset;
+  int32_t right = content_rect.right + offset - 1;
+  int32_t bottom = content_rect.bottom + offset - 1;
+
+  port.data.draw_horizontal_line(left, right, top, 0, top_left_color);
+  port.data.draw_vertical_line(left, top, bottom, 0, top_left_color);
+  port.data.draw_horizontal_line(left, right, bottom, 0, bottom_right_color);
+  port.data.draw_vertical_line(right, top, bottom, 0, bottom_right_color);
+}
+
+static void draw_dbox_frame(CCGrafPort& port, const Rect& content_rect) {
+  constexpr int16_t frame_offset = 6;
+  constexpr int16_t edge_offset = frame_offset - 1;
+  constexpr int16_t shadow_trim = 2;
+
+  // Draw Classic Mac dBoxProc chrome outside the dialog content rect.
+  draw_dbox_frame_line(port, content_rect, frame_offset, 0x000000FF, 0x000000FF);
+  port.data.draw_horizontal_line(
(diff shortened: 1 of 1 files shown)


src/WindowManager.cpp
```diff
@@ -1771,7 +1842,8 @@ WindowPtr WindowManager_CreateNewWindow(int16_t res_id, bool is_dialog, WindowPt
     ref_con = wind.ref_con;
   }
 
-  // If there's a corresponding dctb or wctb, use it to initialize the port's background color. It seems none of the other fields are relevant: they're used for drawing the window frame, but Realmz only uses borderless windows.
+  // If there's a corresponding dctb or wctb, use it to initialize the port's background color.
+  // The other color-table fields describe Toolbox frame chrome; WindowManager draws that separately.
```
danapplegate, line 1846: I'm not sure how much it really matters, but according to "Macintosh Toolbox Essentials," it seems that the `wctb` and `dctb` resources are meant to control the frame color, as well as the text and highlight colors (e.g. pages 4-71 and 6-156). It seems there is some variation in those resources in The Family Jewels. I haven't mapped them out to the specific windows and dialogs that they reference, so I can't really say how much of a visual difference they make. But, if we were to try to implement `draw_dbox_frame` as faithfully as possible, it should take its colors from the `dctb` resource rather than hardcoding the values. Up to you if you want to go that extra mile; I think what you've got here is already a big improvement!

danapplegate (approved):

src/WindowManager.cpp
iSynic: Thanks @danapplegate. I dug into this a little more. Apple defines `dctb`/`wctb` parts `9` and `10` as the light and dark dialog-frame colors, but the Realmz resources I checked in 7.1.2, 8.0.7, and the modern build only contain parts `0–4`.
So it looks like the bevel colors were coming from the Mac OS defaults rather than Realmz’s resources. The tables do contain background, outline, and text-related colors, but they don’t provide a direct replacement for the hardcoded dBox palette here.
I will look into honoring the rest of those fields as a follow-up PR, but it'll be a broader Toolbox color-table implementation.

PR #281, opened 2026-06-29
Support legacy scenario resource sidecars

### LINKED ISSUE #280: Some third-party scenario resource sidecars do not load in the modern port
The modern port expects scenario resource sidecars to be raw `.rsrc` files. That works for some scenarios, but there are at least two other formats in the wild:

- Some `.rsrc` files are AppleDouble containers, where the actual Mac resource fork is stored inside entry `2`.
- Some older Windows-packaged scenarios use `.rsf` sidecars instead of `.rsrc`. Same raw file, different file extension.

For example, I ran into this while trying to add/play Araman's Ring. The scenario folder has a `Scenario.rsrc`, but that file is AppleDouble-wrapped instead of being a raw resource fork. The modern port tries to parse the whole AppleDouble file as a resource fork, which fails and causes a crash.

I also found several scenarios that use `.rsf` files, such as Hax, Lord of the Abyss, The End Worlds, and Wrath of the Mind Lords.

A likely fix is:

- keep `.rsrc` as the preferred modern sidecar
- if a sidecar is AppleDouble, unwrap resource-fork entry `2` before passing it to the resource parser
- treat `.rsf` as a legacy read fallback only; `.rsrc` remains the normal modern sidecar format

iSynic:
Fixes #280.

This fixes the crash I hit when importing some older third-party scenarios, including Araman's Ring.

The scenario resource sidecar exists, but it is not always in the raw `.rsrc` shape the modern port expects. Some `.rsrc` files are AppleDouble containers, and a few old Windows scenario packages use `.rsf` sidecars instead.

This keeps the normal `.rsrc` path, unwraps AppleDouble resource-fork entry `2` when needed, and falls back to `.rsf` only when no `.rsrc` exists.

This does not add new scenario data behavior. It just lets the existing resource manager see the resource fork data those scenarios already shipped with and prevents the crash on import as a result.

### DIFF
--- a/src/ResourceManager.cpp
+++ b/src/ResourceManager.cpp
@@ -3,6 +3,9 @@
 #include <filesystem>
 #include <memory>
 #include <stddef.h>
+#include <stdint.h>
+#include <stdexcept>
+#include <string>
 
 #include "FileManager.hpp"
 #include "MemoryManager.hpp"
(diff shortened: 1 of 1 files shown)


src/ResourceManager.cpp
```diff
@@ -389,17 +449,19 @@ void FSpCreateResFile(const FSSpec* spec, OSType creator, OSType fileType, Scrip
...
+      auto filename = string_for_pstr<64>(spec->name);
+      rm_log.info_f("Failed to load resource file {}", filename);
+      resError = fnfErr;
+      return -1;
+    }
+
     std::string data = phosg::load_file(host_filename);
-    auto rf = std::make_shared<ResourceDASM::ResourceFile>(ResourceDASM::parse_resource_fork(data));
+    auto rf = std::make_shared<ResourceDASM::ResourceFile>(
+        ResourceDASM::parse_resource_fork(resource_fork_data_for_host_file(host_filename, data)));
```
fuzziqersoftware: We should use resource_dasm's AppleSingle/AppleDouble parser instead of reimplementing it. Presumably this codepath would be something like this (maybe with a log message in the catch block):

    std::shared_ptr<ResourceDASM::ResourceFile> rf;
    try {
        rf = std::make_shared<ResourceDASM::ResourceFile>(ResourceDASM::parse_applesingle_appledouble_resource_fork(data));
    } catch (const std::runtime_error& e) {
        rf = std::make_shared<ResourceDASM::ResourceFile>(ResourceDASM::parse_resource_fork(data));
    }

fuzziqersoftware (changes requested):

src/ResourceManager.cpp
iSynic: > We should use resource_dasm's AppleSingle/AppleDouble parser instead of reimplementing it. Presumably this codepath would be something like this (maybe with a log message in the catch block):
> 
> ```
> std::shared_ptr<ResourceDASM::ResourceFile> rf;
> try {
>     rf = std::make_shared<ResourceDASM::ResourceFile>(ResourceDASM::parse_applesingle_appledouble_resource_fork(data));
> } catch (const std::runtime_error& e) {
>     rf = std::make_shared<ResourceDASM::ResourceFile>(ResourceDASM::parse_resource_fork(data));
> }
> ```

I'll take a look, thanks!

Edit: This is addressed.

fuzziqersoftware (approved):

Issue #78, opened 2025-03-13
Clean up ResourceManager error handling conventions


danapplegate:
See this [comment](https://github.com/danapplegate/realmz/pull/75#discussion_r1990654689)

danapplegate:
Resolved by #285

PR #285, opened 2026-07-06
Handle missing resources consistently

iSynic:
Addresses #78.

Realmz had mixed missing-resource behavior in the modern ResourceManager layer. Some missing resources returned `nullptr`, but others threw before original-style callers could handle the missing resource themselves.

That made optional missing resources fatal in places that already had null checks. Trial By Fire exposed this with a Show Pict action that references a missing PICT resource; Classic Mac Realmz skips it, but the current port crashed before `GetPicture()` could return `nullptr`.

This changes `GetResource()` and `GetResourceSizeOnDisk()` to use the Classic-style missing-resource behavior: set `resNotFound` and return `nullptr` / `-1`.

Since some modern wrapper code was implicitly relying on `GetResource()` throwing, this also adds explicit null checks before decoding required resources such as fonts, menus, icons, cursors, controls, dialogs, and windows. Optional resources can now be skipped cleanly, while required resources still fail clearly.

Verified with a Windows build and Trial By Fire. The scenario no longer crashes on the missing picture reference.

### DIFF
--- a/src/ResourceManager.cpp
+++ b/src/ResourceManager.cpp
@@ -271,14 +271,6 @@ public:
       }
     }
 
-    if (type == ResourceDASM::RESOURCE_TYPE_snd) {
-      // TODO: Some action points seem to specify invalid sound resources by id, such as
-      // 0 (which should be the system beep, but which we haven't implemented). If we can't
-      // find the snd resource, we simply return nullptr here, which causes PlaySound to not
-      // play anything.
-      return nullptr;
-    }
-
     std::string type_str = ResourceDASM::string_for_resource_type(type);
     rm_log.info_f("{}:{} not found in any open resource file", type_str, id);
     this->print_chain();
@@ -562,14 +554,16 @@ void UpdateResFile(int16_t refnum) {
 }
 
 Handle GetResource(ResType type, int16_t id) {
-  auto res = rm.get_resource(type, id);
-  if (res != nullptr) {
-    resError = noErr;
-    return res->data_handle;
+  std::shared_ptr<ResourceManager::Resource> res;
+  try {
+    res = rm.get_resource(type, id);
+  } catch (const std::out_of_range&) {
+    resError = resNotFound;
+    return nullptr;
   }
 
-  resError = resNotFound;
-  return nullptr;
+  resError = noErr;
+  return res->data_handle;
 }
 
 Handle Get1Resource(ResType type, int16_t id) {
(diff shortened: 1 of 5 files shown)


chromancer:
I will not review as this seems Dan or Martin's scope and they are requested but I will contribute that running this in my preflight build has no ill effects and it looks fine to me

src/ResourceManager.cpp
```diff
@@ -562,7 +562,12 @@ void UpdateResFile(int16_t refnum) {
 }
 
 Handle GetResource(ResType type, int16_t id) {
-  auto res = rm.get_resource(type, id);
+  std::shared_ptr<ResourceManager::Resource> res;
+  try {
+    res = rm.get_resource(type, id);
+  } catch (const std::out_of_range&) {
+  }
```
danapplegate, line 563: I think this can be cleaned up by moving the error return case below into this `catch` block. So, instead of this empty `catch`, we just do this:

```c++
Handle GetResource(ResType type, int16_t id) {
  std::shared_ptr<ResourceManager::Resource> res;
  try {
    res = rm.get_resource(type, id);
  } catch (const std::out_of_range&) {
    resError = resNotFound;
    return nullptr;
  }

  resError = noErr;
  return res->data_handle;
}
```

We should also remove lines 274-280, the clause that checks `if (type == ResourceDASM::RESOURCE_TYPE_snd)`, to ensure that `ResourceManager::get_resource` always either returns a valid resource pointer or throws. That clause is unnecessary with these changes anyway, since we're updating `GetResource` to return `nullptr` for all missing resources, including `snd`s.

src/ResourceManager.cpp
```diff
@@ -602,7 +607,12 @@ Handle Get1Resource(ResType type, int16_t id) {
 }
 
 int32_t GetResourceSizeOnDisk(Handle data_handle) {
-  auto res = rm.get_resource(data_handle);
+  std::shared_ptr<ResourceManager::Resource> res;
```
danapplegate: The same sort of logic simplification can be done here.

src/ResourceManager.cpp
iSynic: Updated as suggested. `ResourceManager::get_resource()` now consistently returns a valid resource or throws, and `GetResource()` converts a missing resource to `resNotFound/nullptr`. I also removed the old `snd` special case.

src/ResourceManager.cpp
iSynic: Updated this to use the same simplified exception path, returning -1 with `resNotFound` when the handle is missing.

danapplegate (approved):
Thanks!

PR #287, opened 2026-07-06
Guard class 6 monster saves

### LINKED ISSUE #286: type 6 spells are auto saved vs gold carrying monsters
type 6 spells still auto-save vs gold-carrying monsters — savevs() out of bounds read, related to #276 and pointed out by @iSynic there.

iSynic:
Fixes #286.

Follow-up to #276.

#276 fixed the direct class 6 resistance check, but `savevs()` still had the same monster array boundary issue when a class 6 spell reached the normal saving throw path.

Monster save and spell immunity arrays only store classes 0-5. A `savevs(6, monster)` call could read one slot past those arrays instead of treating class 6 as having no matching monster save/immunity entry.

This keeps the existing special `which == 7` aggregate save behavior unchanged, and only guards the normal monster save/immunity lookup to valid stored classes.

Verified with a Windows Debug build.

### DIFF
--- a/src/realmz_orig/misc.c
+++ b/src/realmz_orig/misc.c
@@ -1210,7 +1210,10 @@ short savevs(short which, short who) {
       special /= 6;
       if (temp <= special + spelladjust)
         return (TRUE);
-    } else {
+    /* *** CHANGED FROM ORIGINAL IMPLEMENTATION ***
+     * Monster save and spellimmune arrays only store classes 0-5.
+     */
+    } else if ((which > -1) && (which < 6)) {
       if ((temp <= monster[who - 10].save[which] + spelladjust) || (monster[who - 10].spellimmune[which]))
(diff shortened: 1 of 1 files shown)


chromancer:
After review, this is fixing a symptom and not the disease. This guard fixes the read out of bounds but the indexing is off.

iSynic:
> After review, this is fixing a symptom and not the disease. This guard fixes the read out of bounds but the indexing is off.

That makes sense. Looking back through `resolvespell.c`, I think you’re right: `damagetype` is already treated as 1-based there with `monst.save[damagetype - 1]`, while `savevs()` indexes the monster save/immunity arrays directly with `which`.

I’m going to re-check the mapping before updating the PR.

chromancer:
I already created #288 for that purpose, didn't know you would be so on top of it! Would you like to review?

PR #290, opened 2026-07-09
calculate all equipment bonuses correctly

chromancer:
updated after review comments

This PR addresses the logic of #289. Close #269, close #274, replaces approach of #268 and #275. kudos @afkelsall for finding the brawn double count. I also expand scope to null the global item and prevent the original behavior of passing stale unintended data to the later unguarded checks.

This resolves all bonuses as intended in the manual and what we can recover of the code intent.

fix:
* sum item.damage over the character's equipped items at the top of attack(), using items[].equip flags. No stored stat has this.
* add 5 times sum to att in the shared path after the base to-hit, so armed, unarmed, and animated attackers all get it.
* remove #268's weapon branch add of to-hit, since any equipped weapon is already in the sum.
* remove duplicate damage add at attack.c:71, since weapon damage already comes with character.damage.
* load a null item into the global item

effects:
* enchanted weapons get damage bonus once
* enchanted weapons get to-hit bonus once
* worn items now add to-hit
* unarmed attacks cannot inherit stale item data and use it to pass unguarded checks

The thing I don't like but is consistent: items equipped in the missile/quiver slots count toward melee to-hit. Yes, equipping a bow with enchant added melee damage in orig realmz, and under this PR it also adds melee to-hit. We should discuss this as a later issue; it defies common sense.

testing: built on Mac. Tested with ad hoc logging backwards and forwards.

Issue #289, opened 2026-07-09
Most gear bonuses (damage and to-hit) are added wrong and so is their display


chromancer:
I don't like being verbose but this deserves some explanation.

principle: +3 item, worn or wielded, should add +3 damage and +15% to hit, once each. Manual: "A Longsword +3 adds 3 damage to the total done and is +15% to hit."

original behavior: `wear()` sums every equipped item's bonus, including weapons, into character.damage (wear.c:179). `attack()` adds the weapon's bonus directly (attack.c:71) *and* `character.damage` later (attack.c:373)—enchanted weapons add damage twice and nothing adds to-hit. Realmz manual's Longsword +3 actually adds 6 damage and is +0% to hit.

current state: #268 added the weapon's to-hit inside the weapon branch. this is strictly correct. But we also need something like #275 to add worn-item to-hit. Any correct equipped-items sum already contains the weapon! If instead of replacing, we extend with #275… character.damage also holds brawn's damage half (manual Table 6 lists brawn To Hit and Damage separately), so `att += 5 * character.damage` creates the double count @afkelsall noted.

best illustrated, e.g. Longsword +3 and the issue with brawn, all wrong counts bolded:

| | longsword +3 dmg | longsword +3 hit | worn +3 dmg | worn +3 hit | brawn 20 dmg | brawn 20 hit |
|---|---|---|---|---|---|---|
| orig Realmz | ***+6*** | ***0*** | +3 | ***0*** | +3 | +15 |
| #268 (merged) | ***+6*** | +15 | +3 | ***0*** | +3 | +15 |
| #275 (proposed) | ***+6*** | +15 | +3 | +15 | +3 | ***+30*** |
| intended per manual | +3 | +15 | +3 | +15 | +3 | +15 |

UI has the same issues with "Attack Bonus" and "Defense Bonus". Realmz manual defines both fields as "composite value…all spell effects, conditions and character attributes."

"Attack Bonus" in updatecharinfo.c:37 and showstats.c:60 shows *some* conditions + `damage * 5`. This is the wrong brawn bonus! Doesn't add tohit, and misses conditions confused/blind.

"Defense Bonus" just below misses confused, blind, tangled, and shield-from-hits—all applied by defender-side roll (attack.c:541-563). Further the two files have drifted. (abs() on tangled/hinder in one, not the other, see #210.)

what I see as solution:
[...]

PR #291, opened 2026-07-09
fix attack & defense bonus in UI

chromancer:
updated with review comments

UI half of #289. Depends on #290, should not be merged otherwise. Completes #210.

In both the big character sheet and the half-size panel, we now correct and standardize the display of attack and defense bonuses as the manual says it ought to be.

fix:
* Attack Bonus = tohit + 5 times equipped + full condition list (confused/blind added), done exactly as attack() does under #290
* Defense Bonus includes its missing conditions: confused, blind, tangled, shield-from-hits
* abs() on condition values in updatecharinfo.c to match. Permanent conditions store negative values. #210 fixed showstats and missed this file.
* detect and add double to-hit of penetration weapons
* update file to use c[charselectnew] instead of characterl, which is stale and not refreshed here (original bug)

effects:
* both screens agree with each other and with logic
* Attack Bonus changes for every character.

note I stash the global item around the sum here. There's a lot of paths to audit.

Things I don't like: the equipped sum is now hand-copied in three places. I did not want to burden this with adding game logic functions for which there is no clean precedent but if I was writing the game from scratch…

testing: built on Mac. Values verified by hand, screens agree, and are valid against the new values from #290.

### DIFF
--- a/src/realmz_orig/showstats.c
+++ b/src/realmz_orig/showstats.c
@@ -3,7 +3,8 @@
 
 /***************************** ShowStats ********************************/
 void ShowStats(short showprestige) {
-  short t, conditionindex = 0;
+  short t, conditionindex = 0, equipped;
+  struct itemattr saveditem;
   int32_t prestige = 0;
   char special[50], specialindex;
 
@@ -44,24 +45,45 @@ void ShowStats(short showprestige) {
   DialogNum(23, characterl.co + characterl.magco);
   DialogNum(24, characterl.lu + characterl.maglu);
 
+  /* *** CHANGED FROM ORIGINAL IMPLEMENTATION ***
+  * NOTE(chromancer): Original displayed attack bonus (to-hit) as "(some) conditions + character.damage * 5",
+  * not accurate to game combat logic or intended behavior. Following original intent, we:
+  * Calculate the attack bonus from conditions, base to-hit, then 5 * equipped (double for penetration)
+  * Complete the defense bonus to match the manual's "all spell effects, conditions".
+  */
+  saveditem = item;
+  equipped = 0;
+  for (t = 0; (t < characterl.numitems) && (t < 30); t++)
+    if (characterl.items[t].equip) {
+      loaditem(characterl.items[t].id);
+      equipped += item.damage;
+      if (item.sp1 == 121)
+        equipped += item.damage; // Double to-hit for penetration
+    }
+  item = saveditem;
+
   temp = 0;
   if (characterl.condition[COND_STRONG])
     temp += 15; /**** strong ***/
   if (characterl.condition[COND_SLOW])
     temp -= 15; /**** slow ***/
+  if (characterl.condition[COND_CONFUSED])
+    temp -= 10;
+  if (characterl.condition[COND_BLIND])
+    temp -= 15;
   if (characterl.condition[COND_MAGIC_AURA])
-    temp += 5; /**** bless ***/
+    temp += 5;
   if (characterl.condition[COND_CURSED])
     temp -= 5; /**** curse ***/
   if (characterl.condition[COND_TANGLED])
     temp -= abs(characterl.condition[COND_TANGLED]); /*** tangled ***/
   if (characterl.condition[COND_HINDERED_ATTACKS])
     temp -= abs(characterl.condition[COND_HINDERED_ATTACKS]); /*** Hinder atk ***/
-  temp += (characterl.damage * 5); /*** Hinder atk ***/
+  temp += characterl.tohit + (5 * equipped);
 
   if (temp > 99)
     TextSize(16);
-  DialogNum(25, temp); /*** attack bonus ****/
+  DialogNum(25, temp); // Total displayed to-hit bonus (UI label "Attack Bonus").
   TextSize(20);
 
   temp = 0;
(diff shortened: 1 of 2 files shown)


src/realmz_orig/showstats.c
```diff
@@ -44,11 +45,30 @@ void ShowStats(short showprestige) {
...
+      equipped += item.damage;
+    }
+  item = saveditem;
+
   temp = 0;
   if (characterl.condition[COND_STRONG])
     temp += 15; /**** strong ***/
   if (characterl.condition[COND_SLOW])
     temp -= 15; /**** slow ***/
+  if (characterl.condition[COND_CONFUSED])
```
afkelsall, line 70: You found, missing modifiers? :P Nice

src/realmz_orig/showstats.c
```diff
@@ -44,11 +45,30 @@ void ShowStats(short showprestige) {
...
+    temp -= 10;
```
afkelsall, line 71: Actually, incoming nit, but since the rest of the modifiers comment next to the modifier value, can you keep this consistent here? Repeat for same things below - I count 6 sites needing a comment I think.
```suggestion
    temp -= 10; /**** confused ***/
```

src/realmz_orig/showstats.c
```diff
@@ -57,7 +77,7 @@ void ShowStats(short showprestige) {
     temp -= abs(characterl.condition[COND_TANGLED]); /*** tangled ***/
   if (characterl.condition[COND_HINDERED_ATTACKS])
     temp -= abs(characterl.condition[COND_HINDERED_ATTACKS]); /*** Hinder atk ***/
-  temp += (characterl.damage * 5); /*** Hinder atk ***/
+  temp += characterl.tohit + (5 * equipped);
```
afkelsall, line 82: I think this is the bit that avoids the brawn double count on the stat sheet.

src/realmz_orig/updatecharinfo.c
```diff
@@ -22,21 +23,40 @@ void updatecharinfo(void) {
...
+    }
+  item = saveditem;
 
   temp = 0;
   if (characterl.condition[COND_STRONG])
     temp += 15; /**** strong ***/
   if (characterl.condition[COND_SLOW])
     temp -= 15; /**** slow ***/
+  if (characterl.condition[COND_CONFUSED])
+    temp -= 10;
```
afkelsall, line 50: I think keep all these consistent too with the little comments in the same style.

src/realmz_orig/showstats.c
afkelsall: Ok I think still keep it consistent, but here's why the original version needed the comments :) 

Image

src/realmz_orig/updatecharinfo.c
```diff
@@ -22,21 +23,40 @@ void updatecharinfo(void) {
...
+  /* *** CHANGED FROM ORIGINAL IMPLEMENTATION ***
+  * NOTE(chromancer): Original displayed attack bonus (to-hit) as "(some) conditions + character.damage * 5",
+  * not accurate to game combat logic or intended behavior. Following original intent, we:
+  * Calculate the attack bonus from conditions, base to-hit, then 5 * equipped.
+  * Add absolute values for conditions that use negative values to indicate permanent status.
+  * Complete the defense bonus to match the manual's "all spell effects, conditions".
+  */
+  saveditem = item;
+  equipped = 0;
+  for (t = 0; (t < characterl.numitems) && (t < 30); t++)
```
afkelsall: How would you like to fix another existing defect, one that is actually stopping your "fixed" code from working correctly still...

You're going to see it right away. The showstats.c file uses `characterl` *throughout* the file. The `updatecharinfo.c` file uses `c[charselectnew]` instead, except when it comes to calculating attack/defense bonus, where it uses a non-updated `characterl`. These should all be `c[charselectnew]` :P

afkelsall (commented):
.

src/realmz_orig/showstats.c
```diff
@@ -44,11 +45,30 @@ void ShowStats(short showprestige) {
...
+  * not accurate to game combat logic or intended behavior. Following original intent, we:
+  * Calculate the attack bonus from conditions, base to-hit, then 5 * equipped.
+  * Complete the defense bonus to match the manual's "all spell effects, conditions".
+  */
+  saveditem = item;
+  equipped = 0;
+  for (t = 0; (t < characterl.numitems) && (t < 30); t++)
+    if (characterl.items[t].equip) {
+      loaditem(characterl.items[t].id);
+      equipped += item.damage;
```
afkelsall, line 59: I don't know how you want to handle this, but getting the displays to represent actual Attack Bonus is still missing a value, and given we're trying to get the field to represent the actual attack bonus, I think it's worth adding here.

item.sp1 == 121 aka piercing weapons have double to hit impact and to my knowledge is not handled here. I'd suggest putting the piercing check it in the item loop, but acknowledging it makes the loops in these two files different from the one in attack.c. And I'd prefer to leave the attack.c code as is where possible.

chromancer: You're right. The PR should do what it says it does. We can refactor later if there's some agreement on cleaning up all the copies of the formula

src/realmz_orig/updatecharinfo.c
chromancer: Good catch. Let's fix it

src/realmz_orig/showstats.c
chromancer: obvious once you just actually put your eyes on attack.c and then look over here. I have a pretty good idea about why it was missed too. How could anybody remember what to check for and what not to when these were just a list of indices?

src/realmz_orig/showstats.c
chromancer: Yeah, damage never should have been here.

src/realmz_orig/showstats.c
chromancer: Right. These comments were added to make the section readable because these were originally just indices to the condition array.

There's a couple of these that are obviously wrong: there is no condition called Bless. There _used_ to be; it's Magic Aura. 2nd and 3rd "Hinder atk" is obviously a copy/paste error. "attack bonus" and "defense bonus" are unclear themselves, since both of those names collide with other real things (COND_ATTACK_BONUS, which is a bonus to damage, and COND_DEFENSE_BONUS, which is called defense bonus but is actually just an input to the Defense Bonus). I think those should be corrected by removal where they are superfluous to the enum or replacement where merited

My position has been that I don't touch og comments (or source!) unless I'm rewriting the line and the line is wrong.

I don't think we should _add_ comments which imitate the og realmz house style, especially not when we have the clear enums. I think someone coming behind me would, unless they're doing git archaeology, be deceived into thinking those lines and comments were untouched since 199X. I am open to a counter-argument.

afkelsall (commented):
FWIW I'm happy to approve this PR, but my invitation has expired.

src/realmz_orig/updatecharinfo.c
```diff
@@ -6,7 +6,8 @@ void updatecharinfo(void) {
   int enable_recomposite = WindowManager_SetEnableRecomposite(0);
 
   GrafPtr oldport;
-  short temp, conditionindex = 0;
+  short temp, conditionindex = 0, equipped;
```
iSynic, line 9: Small nit pick... This loop currently uses the shared global `t`. None of the current callers seem to rely on its previous value, but could we declare `t` locally with the other shorts here? That would keep `updatecharinfo()` from acquiring an unrelated global side effect and would match `ShowStats()`.

short temp, conditionindex = 0, equipped, t;

Non blocking so approved.

iSynic (approved):
global `t` is nonblocking.

src/realmz_orig/updatecharinfo.c
chromancer: `t` is a scratchpad common in this era for this purpose (see `tt` and `ttt`)

not something we'd do nowadays but in this case I am following the idiom for clarity's sake and because I checked it to be harmless. `showstats.c` is odd; orig `updatecharinfo` runs display loops on `t`

Thanks for review pal!

Issue #292, opened 2026-07-11
Inconsistent handling of stats between character generation and later modification


Orthotope:
In most cases, changing stats by spell effect (resolvespell.c, effect 66) or aging does not have the same effect as at character creation.

Brawn: should provide modifiers to special abilities. Only calculated at character creation, increasing later has no effect. Bonuses to damage and to-hit are recalculated when changed by spell or aging.
Knowledge, Judgment: at creation, magic resistance is calculated as (knowledge + judgment) / 10, scaled by caste multiplier. Later increases will add one point (scaled by caste multiplier) if the new stat value is 16 or higher.
Agility: increases armor rating by 2 for every point above 14. At creation only, gives 2% missile dodge chance per point. Like Brawn, modifiers to special abilities are only calculated at character creation.
Vitality: increases all DR Vs. by 5% when modified to 19 or higher; this is not included at character creation.

Changes to stats from aging (other than Brawn) do not trigger any recalculations.

In resolvespell.c, the indexed stat is increased by 1. In the case of increased stamina or spell points (effect size 10 and 11), this writes to the Sorcerer spell list, giving Flame Hands and Free Fall respectively.

PR #293, opened 2026-07-11
Consistent handling of stat changes

Orthotope:
During character creation, high knowledge/judgment/vitality will grant the same resistance bonuses you would get by increasing those stats to the same level later.

Modifying stats, whether by spell effect/potion or by aging, will now update derived stats (to-hit, damage, armor rating, missile dodge chance, strength- and dex-based special abilities, magic resistance, DR Vs) accordingly.

Magic strength from equipped items is included in determining the above derived stats.

### DIFF
--- a/src/realmz_orig/age.c
+++ b/src/realmz_orig/age.c
@@ -130,14 +130,14 @@ void applyage(short raceid, short agegroup, short direction) {
 
   characterl.currentagegroup = agegroup;
 
-  strength(characterl.st);
+  strength(characterl.st + characterl.magst);
   characterl.tohit -= temp;
   characterl.damage -= damage;
   updatespec(2);
   
   characterl.st += direction * races.agechange[agegroup - 1][0];
 
-  strength(characterl.st);
+  strength(characterl.st + characterl.magst);
   characterl.tohit += temp;
   characterl.damage += damage;
   updatespec(3);
--- a/src/realmz_orig/age.c
+++ b/src/realmz_orig/age.c
@@ -123,24 +123,60 @@ void showageupdate(short who, short agegroup, short backup) {
 void applyage(short raceid, short agegroup, short direction) {
   short t;
 
-  loadprofile(raceid, 0);
+  /* *** CHANGED FROM ORIGINAL IMPLEMENTATION ***
+   * NOTE(orthotope): recalculate stats based on changes
+   */
+  loadprofile(raceid, characterl.caste);
 
   characterl.currentagegroup = agegroup;
 
   strength(characterl.st);
   characterl.tohit -= temp;
   characterl.damage -= damage;
-
+  updatespec(2);
+  
   characterl.st += direction * races.agechange[agegroup - 1][0];
 
   strength(characterl.st);
   characterl.tohit += temp;
   characterl.damage += damage;
+  updatespec(3);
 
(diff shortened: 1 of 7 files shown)


iSynic:
I did a deeper fixture pass on this because stat changes touch several cached character fields at once, and I was having trouble validating all of the paths independently. I put the version I tested here as a reference:

https://github.com/iSynic/realmz/tree/test-consistent-stat-updates

This is not meant as a competing PR. I just wanted a runnable version so I could test the cases I was concerned about.

The main issue I found in the current implementation is that `updatespec()` operates on the global `characterl`, while the new calls from `resolvespell()`, `wear()`, and `removeitem()` are modifying characters in `c[t]` or `c[character]`. Those calls can then update the wrong character's `spec[]` values.

I also found that the new Brawn and Agility modifier loops iterate 14 entries, while `stmodif` and `demodif` only have 12 rows. The 14-entry count is correct for the race/caste special-ability tables, but using it for these modifier tables reads past their bounds.

The version I tested passes the affected character directly to the helper. It removes the old derived modifiers, changes the underlying attribute, and then reapplies them through one shared path. The modifier-table loops are limited to the 12 rows that actually exist, while the normal race/caste special-ability handling remains at 14.

I tested permanent Brawn, Knowledge, Judgment, Agility, and Vitality changes; stamina and spell-point increases; level-up; Magic Strength equipment round trips; and aging/youth transitions. The derived values reversed cleanly when returning to the original age group, and changing one party member did not affect the others.

The runtime testing also found a related Brawn case and I wasn't sure if it's related to #292... Changing base Brawn through effect `66` or aging does not recalculate `loadmax`. Brawn equipment does, because `wear()` and `removeitem()` call `movecalc()`. I left that out of this branch because `applyage()` also changes movement directly, so I think carrying capacity and the age-related movement calculations need to be looked at separately.

Orthotope:
Thanks for taking a look at this. I'm not particularly attached to the PR itself; if someone else comes up with a better implementation, go ahead and use that.

The loop count is based on the existing code for switch case 1 (new character generation). Should that also be adjusted to not read outside of the arrays, or should the arrays be extended to 14 rows? The new ones would be Read Scroll (no longer used) and Turn Undead.

iSynic:
I checked the manual and I think it clears up both questions.

The race/caste special-ability arrays really are 14 entries, so general initialization and level-up should use 14. The Brawn and Agility modifier tables are intentionally only 12 rows - Read Scroll and Turn Undead don't have attribute-modifier rows.

The manual also says those Brawn/Agility adjustments apply only to initial attribute scores, and that magically altered attributes do not award additional bonuses or penalties. So I need to revise the branch I linked since later stat changes should update ordinary derived fields, but shouldn't remove and reapply the special-action modifier tables.

That leaves 14 for race/caste ability initialization and level-up, and 12 for `stmodif`/`demodif`. I wouldn't extend the modifier tables.

chromancer:
Which manual? Divinity manual appears to be purely descriptive whereas Realmz manual appears more prescriptive

I'm no longer inclined to take the Divinity manual as evidence of intent given some of the janky stuff documented in it

iSynic:
Realmz Manual 8.0. 

image

And for what it's worth, I 100% agree with you about the Divinity Manual.

PR #295, opened 2026-07-12
Fix on-portrait anims (spells, pots, selection) by adding missing leftshifts

chromancer:
issue: portrait frame feedback and spell/potion flash animations draw visibly left of the their intended locations.

cause: When realmz was changed from a hand-tailored 640x480 layout to 800x600, various UI elements were shifted in location from their hardcoded locations by (among other ways) applying `leftshift` to move elements exactly 160px over.

I suspect the source we inherited was never really finished or they just got lazy because several draws are missing the necessary leftshifts. This is the result:

Screenshot 2026-07-12 at 2 25 40 AM

Note the red ball in the item rect which is a draw artifact of an animation which should play over a portrait which exists in a window that is z-under the short sheet to the right (i.e., it should not be seen)

fix:
* add + leftshift at four sites (buttonchoice.c x2, combatchoice, spelleffect)
* that's it

effects:
* pressing a portrait flashes the pressed portrait instead of drawing the frame to its left
* spell/potion anims don't cause artifacts

notes: testing and verification will show mostly nothing, just the absence of incorrect draws in some situations. our current render arch is (I think) not capable of capturing the selection frame anims whether they exist or not. I have a marvelous proof of this which this margin is too small to contain.

### DIFF
--- a/src/realmz_orig/buttonchoice.c
+++ b/src/realmz_orig/buttonchoice.c
@@ -501,7 +501,10 @@ moveon:
     charselectnew = point.v / 50;
     itemRect.top = charselectnew * 50;
     itemRect.bottom = itemRect.top + 50;
-    itemRect.left = 330;
+    /* *** CHANGED FROM ORIGINAL IMPLEMENTATION ***
+     * NOTE(chromancer): Added missing leftshifts where noted inline.
+     * These caused portrait feedback anims to be drawn too far left. */
+    itemRect.left = 330 + leftshift; // Missing leftshift.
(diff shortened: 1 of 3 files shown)


iSynic (approved):
I tested this locally with the PR applied to current main, using Oil of Resistance from the Items screen and a few other spell effect consumables and spells.

The animation is moved back over the character panel (under the items screen) where it belongs, matching the behavior in 7.1.2. I also checked the related portrait positioning changes, and everything looks correct to me.

Approved.

PR #297, opened 2026-07-12
Fix spell info lights

### LINKED ISSUE #296: Incorrect handling of spell damage type lights
In castspell.c:478, the order of operations is incorrect:
`type = abs(spellinfo.damagetype - 1);`
The intention here is to map the damage types heat through magic (normally 1 through 6) into the range 0-5. For most spells with non-negative damage types, the result is correct. However, for damage type 0 (charm, such as Charm Foe), `type` is calculated to be 1, and so the Cold light is active when it should be off. Negative damage types are used to indicate a save vs caster, almost exclusively used for Mental spells (damage type -5). For these, `type` evaluates to 6 instead of 4, so the Mental light is off when it should be on.

Fix is to move the -1 offset outside the abs() call.

chromancer:
issue: Closes #296. The cold light is on for charm `damagetype` and the mental light doesn't come on when it should because realmz needs to take the absolute value before the index offset.

fix:
* move parenthesis

effects: lights are now visibly correct.

testing: Built on Mac and verified with Soul Bind (mental light now on) and Charm Foe (cold light stays off)

### DIFF
--- a/src/realmz_orig/castspell.c
+++ b/src/realmz_orig/castspell.c
@@ -475,7 +475,11 @@ back:
 
         updateinfo:
 
-          type = abs(spellinfo.damagetype - 1);
+          /* *** CHANGED FROM ORIGINAL IMPLEMENTATION ***
+           * NOTE(chromancer): Original spellinfo panel had wrong order of ops.
+           * Corrected to use proper index for charm and mental types.
+           */
+          type = abs(spellinfo.damagetype) - 1;
           DrawPicture(non, &typerect);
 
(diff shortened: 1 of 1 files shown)


iSynic (approved):
Looks good to me. I checked the mapping across the full range used by the six lights.

With this change, `0` correctly leaves all damage-type lights off, positive values `1` through `6` map to lights `0` through `5`, and negative save-vs-caster values map to the same lights after the sign is removed. Values outside that range are still ignored by the existing bounds check.

I also built and ran the PR on Windows without issue.

PR #298, opened 2026-07-14
Fix EDIT_TEXT caret rendering

### LINKED ISSUE #166: Character generation screen: No blinking text entry cursor is visible in Name entry area.
Gives affordance you can't enter text. Typing does still work however.

Image

Image

iSynic:
Fixes #166.

The current `EDIT_TEXT` implementation always draws a solid caret while the field has focus. It also draws the caret one pixel outside the field's erased area, which can leave dots or lines behind as text is entered.

image

Realmz 7.1.2 uses the normal TextEdit idle behavior to blink the caret. This adds the equivalent behavior to the port, keeps the caret visible while typing, and confines it to the editable field so it erases cleanly.

Caret idle updates are handled when the event queue is empty.

Tested with the Add Divinity Scenario text entry dialog and compared with 7.1.2.

image

### DIFF
--- a/src/EventManager.cpp
+++ b/src/EventManager.cpp
@@ -324,6 +324,10 @@ public:
   EventRecord get_next_event(uint32_t wait_ms) {
     this->enqueue_pending_events(wait_ms);
     if (this->event_queue.empty()) {
+      auto window = WindowManager::instance().front_window();
+      if (window) {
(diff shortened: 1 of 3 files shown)


danapplegate (approved):
Nice work, thanks! Tested it out on Mac, looks great.

Issue #259, opened 2026-06-22
Viewing a map could/should center in the game/map area


afkelsall:
When viewing a map it sits in the top left of the game area and can be difficult to distinguish from the rest of the game still displaying underneath.

This looks like like an artifact from the 640x400 -> 800x600 migration without updating the displaying of the map.

I think it could either center and black out the background while it's up. An alternative, though I don't think is desired is to enlarge the map and display more but I'm not even sure if maps work like this.

I'd opt for centered and black out the rest of the view area.

<img width="1202" height="952" alt="Image" src="https://github.com/user-attachments/assets/b1ddb129-9ebd-4e31-840d-9270ec45e375" />

jpetrie:
Centering and blacking out is probably the right call - expanding the size of the area visible in the map might:

- lead to problems around border of the map, where we may need to black out tiles that would be out of range anyhow
- reveal more of an area than intended, making it easier to work out where a map location is

I think an ideal solution to this would be to center the map in the gameplay area, blacking out the extra space, and moving the popup text messages out of the way a bit more - the map description can probably get moved all the way down and the "Click Mouse" panel can go completely over the character list.

iSynic:
> Centering and blacking out is probably the right call - expanding the size of the area visible in the map might:
> 
>     * lead to problems around border of the map, where we may need to black out tiles that would be out of range anyhow
> 
>     * reveal more of an area than intended, making it easier to work out where a map location is
> 
> 
> I think an ideal solution to this would be to center the map in the gameplay area, blacking out the extra space, and moving the popup text messages out of the way a bit more - the map description can probably get moved all the way down and the "Click Mouse" panel can go completely over the character list.

Is there any reason it couldn't just be scaled to suit the current viewport? 

> Centering and blacking out is probably the right call - expanding the size of the area visible in the map might:
> 
>     * lead to problems around border of the map, where we may need to black out tiles that would be out of range anyhow
> 
>     * reveal more of an area than intended, making it easier to work out where a map location is
> 
> 
> I think an ideal solution to this would be to center the map in the gameplay area, blacking out the extra space, and moving the popup text messages out of the way a bit more - the map description can probably get moved all the way down and the "Click Mouse" panel can go completely over the character list.

There are a few possible approaches here, depending on how far we want to go with restoring the map presentation after the 800x600 conversion.

The conservative option would be to keep the generated map at its original 320x320 presentation and just center it in the larger viewport.

[...]

PR #300, opened 2026-07-15
Center map content in the enlarged viewport

iSynic:
Addresses part of #259.

The 800x600 conversion left maps authored for a 320x320 display area drawing from the top-left of the enlarged viewport.

This keeps generated land maps at their original 320x320 coverage and centers them in the game viewport. Overlay icons and the party marker remain aligned with the map, and the small amount of tile spill that the original 320x320 port would have clipped is hidden again.

Picture-backed maps remain at their native size. Maps with an authored destination rectangle still use it; maps without one are centered.

This does not change map data, zoom, or the number of tiles displayed.

image

image

### DIFF
--- a/src/realmz_orig/mapstuff.c
+++ b/src/realmz_orig/mapstuff.c
@@ -63,7 +63,7 @@ void showmap(short mapnumber) {
   PicHandle picture = NIL;
   Rect temprect, margin;
   Rect maprect;
-  Boolean centeredgeneratedmap = FALSE;
+  Boolean centeredmapcontent = FALSE;
   /* *** END CHANGES *** */
   short oldview, t, tt, temp, tempisdung;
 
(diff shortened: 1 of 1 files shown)


src/realmz_orig/mapstuff.c
```diff
@@ -80,18 +87,43 @@ void showmap(short mapnumber) {
     movie(themap.show, 129, 0);
```
chromancer, line 95: I see an opportunity to hang the whole thing off of the maprect, set up once above the branches, e.g.,

```c
  SetRect(&maprect, 0, 0, 320, 320);
  OffsetRect(&maprect, lookrect.left + ((lookrect.right - lookrect.left) - 320) / 2,
      lookrect.top + ((lookrect.bottom - lookrect.top) - 320) / 2);
```

Start the tile loop at the maprect origin
DrawPicture already scales into it properly
#302 can copy right into it
icons and marker offset directly from maprect origin
one crop for land and dungeon maps

Ought to unify #300 and #302 nicely. These will have to rebase anyway.

chromancer (changes requested):

src/realmz_orig/mapstuff.c
iSynic: This should be implemented now and an updated #302 will follow shortly.

chromancer (approved):
All discussion addressed. Verified this looks good in Mac build. Approved.

PR #301, opened 2026-07-15
Restore map note presentation

iSynic:
Addresses part of #259.

The enlarged layout was still using the original map-note dialog and positioning both it and the `Click Mouse` prompt with the old screen coordinates.

This uses the existing large-screen dialog, places it over the lower-left message panel, and restores the yellow outline around the note text. The `Click Mouse` prompt is moved over the right-side status panel so it no longer obscures the map.

image

### DIFF
--- a/src/realmz_orig/mapstuff.c
+++ b/src/realmz_orig/mapstuff.c
@@ -222,7 +222,11 @@ void showmap(short mapnumber) {
   else
     loadland(dunglevel, TRUE);
 
-  flashmessage((StringPtr) "Click Mouse", 350, 100, 0, 30005);
+  /* *** CHANGED FROM ORIGINAL IMPLEMENTATION ***
+   * NOTE(iSynic): Keep the map prompt over the status panel in the enlarged layout.
+   */
+  flashmessage((StringPtr) "Click Mouse", 350 + leftshift, 100, 0, 30005);
+  /* *** END CHANGES *** */
 
   xy(0);
   DisposeDialog(show);
--- a/src/realmz_orig/mapstuff.c
+++ b/src/realmz_orig/mapstuff.c
@@ -196,11 +196,16 @@ void showmap(short mapnumber) {
     }
   }
 
-  show = GetNewDialog(169, 0L, (WindowPtr)-1L);
+  /* *** CHANGED FROM ORIGINAL IMPLEMENTATION ***
+   * NOTE(iSynic): Use the large-screen map note dialog and match the spell
+   * description panel placement.
+   */
+  show = GetNewDialog(169 + (1000 * screensize), 0L, (WindowPtr)-1L);
   SetPortDialogPort(show);
   BackPixPat(base);
   TextFont(defaultfont);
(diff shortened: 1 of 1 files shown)


src/realmz_orig/mapstuff.c
```diff
@@ -209,6 +226,11 @@ void showmap(short mapnumber) {
 
   DrawDialog(show);
   MyrPascalDiStr(2, themap.note);
+  GetDialogItem(show, 2, &itemType, &itemHandle, &itemRect);
+  InsetRect(&itemRect, -2, -2);
+  ForeColor(yellowColor);
+  FrameRect(&itemRect);
```
chromancer: Something doesn't look right here… We shouldn't need to manually draw a frame. I think this is a port layer problem

src/realmz_orig/mapstuff.c
```diff
@@ -196,11 +196,28 @@ void showmap(short mapnumber) {
...
+    Rect bounds;
+    short noteWidth, noteHeight, noteLeft;
+
+    GetPortBounds(GetWindowPort(show), &bounds);
+    noteWidth = bounds.right - bounds.left;
+    noteHeight = bounds.bottom - bounds.top;
+    noteLeft = 0;
+    if (noteWidth < buttons.left)
+      noteLeft = (buttons.left - noteWidth) / 2;
+    MoveWindow(GetDialogWindow(show), GlobalLeft + noteLeft, GlobalTop + info.bottom - noteHeight - 1, FALSE);
```
chromancer: I get what you're doing here but I think the 800x600 conversion really confused layout. Check this out, this is what it looks like under the PR.


Image


However if we do, instead of the entire if/else block,

```c
MoveWindow(GetDialogWindow(show), GlobalLeft - 1, GlobalTop + 321 + downshift, FALSE);
```

we get this:


Image


Check out castspell.c at 171 where they got it right. Since the DLOG is only width modified, this works for screensize true or false.

src/realmz_orig/mapstuff.c
chromancer: *Macintosh Toolbox Essentials* on editText items "The Dialog Manager uses the QuickDraw procedure FrameRect to draw a rectangle three pixels outside the display rectangle"

So our issue is that our port layer EDIT_TEXT case needs to draw the correct frame, but it doesn't. Fixing on this layer would yeah expand the scope of the PR, but it restores the correct border to all dialog items of this type.

We could either do that as a follow-up or do it here, simple enough, but I don't think we want to draw it manually when we'd have to do the same elsewhere

chromancer (changes requested):
My apologies on the late review. I wrote these comments on the 28th but left them pending. I will re-review asap or ping me to discuss.

src/realmz_orig/mapstuff.c
iSynic: Rebased and narrowed to suit #315 . I think this is in line with your suggestions now.

chromancer (approved):
Builds clean and looks good!

PR #302, opened 2026-07-15
Center dungeon maps in the enlarged viewport

iSynic:
Addresses part of #259.

Dungeon maps are still rendered into their original 320x320 buffer, but the enlarged layout was leaving that buffer at the top-left of the viewport.

This clears the enlarged viewport and copies the existing dungeon map into its center without scaling it or changing the area shown.

image
image

### DIFF
--- a/src/realmz_orig/mapstuff.c
+++ b/src/realmz_orig/mapstuff.c
@@ -56,9 +56,8 @@ void fastplotmap(short id, Rect destrect) {
 void showmap(short mapnumber) {
   FILE* fp = NULL;
   DialogRef show;
-  Boolean tag = FALSE;
   /* *** CHANGED FROM ORIGINAL IMPLEMENTATION ***
-   * NOTE(iSynic): Track the centered native-size map presentation.
+   * NOTE(iSynic): Track the centered Classic 320x320 map presentation.
    */
   PicHandle picture = NIL;
   Rect temprect, margin;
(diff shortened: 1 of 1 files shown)


chromancer:
I'm guessing this is not the correct screenshot?

iSynic:
You were right, wrong screenshot. Fixed.

iSynic:
After a rebase, this should be ready to take advantage of #300 (once it's merged).

src/realmz_orig/mapstuff.c
```diff
@@ -56,8 +56,14 @@ void fastplotmap(short id, Rect destrect) {
 void showmap(short mapnumber) {
   FILE* fp = NULL;
   DialogRef show;
-  Boolean tag = FALSE;
-  Rect temprect;
+  /* *** CHANGED FROM ORIGINAL IMPLEMENTATION ***
+   * NOTE(iSynic): Track the centered native-size map presentation.
```
chromancer: You may want to update the comment here.

chromancer (commented):
Ready to approve after update stale comment and rebase

src/realmz_orig/mapstuff.c
iSynic: Thanks, done.

chromancer (approved):
Looks good, builds clean. Approved.

PR #303, opened 2026-07-15
Restore 3D dungeon view presentation

iSynic:
Addresses part of #259.

The enlarged layout left two presentation differences in the 3D dungeon view.

The SDL QuickDraw implementation rendered `FillRect()` as a solid color even when Realmz selected `patCopy` with `PenPixPat()`. This caused the original blue patterned background to be replaced by a solid fill.

The 800x600 layout also offset the compass by half of the added viewport height, leaving it floating above the lower interface instead of sitting directly above it.

This adds patterned `FillRect()` support using the existing PixPat decoder and anchors the bottom of the compass to the bottom of the gameplay viewport. Other fill modes, the compass size, and the centered 3D view are unchanged.

I submitted the two changes as separate commits in the hope they'd be easier to review.

image

### DIFF
--- a/src/QuickDraw.cpp
+++ b/src/QuickDraw.cpp
@@ -49,6 +49,8 @@ static std::unordered_map<int16_t, ResourceDASM::BitmapFontRenderer> bm_renderer
 
 std::unordered_set<const CCGrafPort*> CCGrafPort::all_ports;
 
+static phosg::ImageRGB888 reference_image_for_ppat(PixPatHandle ppat);
+
 CCGrafPort* CCGrafPort::as_port(void* ptr) {
   auto* port_ptr = reinterpret_cast<CCGrafPort*>(ptr);
   return all_ports.count(port_ptr) ? port_ptr : nullptr;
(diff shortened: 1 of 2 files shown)


src/QuickDraw.cpp
```diff
@@ -114,6 +116,24 @@ void CCGrafPort::erase_rect(const Rect& r) {
 }
 
 void CCGrafPort::fill_rect(const Rect& r) {
+  if (this->pnMode == 0x08) { // patCopy
```
chromancer, line 119: There's no reason this function has to be named faithfully to Color QD but it did confuse me for a hot minute. `FillRect`, `FillCRect` specifically ignore the pen pattern so why would our port layer need to implement it? Because this isn't them, it's `PaintRect`.

chromancer:
Pattern/compass looks as faithful as possible to the original presentation. Matches what we should expect from Inside Macintosh: Imaging with QuickDraw.

I think we ought to consider renaming `fill_rect` to `paint_rect` and putting unconditional fills on `draw_rect` renamed to `fill_rect` or what have you. It would just be cleaner. `draw_rect` is dead.

That's really irrelevant to this PR, which is nice and tight and deserves to get in the our next beta. Approved.

chromancer (approved):

PR #304, opened 2026-07-19
Stop dropping treasure screen clicks made during the item animation

afkelsall:
Clicking items in quick succession on the post-fight treasure screen silently loses some of them: the click registers, nothing happens, and the item has to be clicked again.

Why:
- `textbox` ends with an unconditional `FlushEvents(everyEvent, 0)`, which discards everything queued at that moment.
- Only the `click` path needs it, to drop the remnants of the click that dismissed the message. The passive path (`click` FALSE) draws a message and returns, so the flush there just destroys unrelated input.
- The treasure screen item hover preview goes through that passive path (`quickinfo` -> `textbox`). Redrawing the preview for a cell therefore flushes the queue.
- A click made while the previous item's keep animation is running is queued, then destroyed by the next hover redraw before the event loop can dequeue it, so that item is never taken. The animation blocks for roughly 100 to 250 ms, which is the window where this happens.
- Original game logic (`src/realmz_orig/`), not the SDL shim, and the same defect on macOS and Windows.

How:
- Guard the trailing flush with `if (click)`, so only the modal dismissal path clears the queue.
- Verified with event logging: before the change, queued `mouseDown` events on the treasure screen were cleared without ever being dequeued; after it, clicks made during the animation survive and are processed.
- Touches only shared game code, so it behaves the same on macOS and Windows.

### DIFF
--- a/src/realmz_orig/textbox-time.c
+++ b/src/realmz_orig/textbox-time.c
@@ -123,7 +123,15 @@ void textbox(short class, short index, short click, short different, Rect newrec
   }
 out:
 
-  FlushEvents(everyEvent, 0);
+  /* *** CHANGED FROM ORIGINAL IMPLEMENTATION ***
+   * NOTE(afkelsall): This flush ran unconditionally, so the passive path
+   * (click FALSE, used for messages drawn without waiting for a dismissal)
+   * also discarded input that had arrived while the game was busy drawing.
+   * Only the click path needs the flush, to drop the remnants of the
+   * dismissing click. */
+  if (click)
+    FlushEvents(everyEvent, 0);
+  /* *** END CHANGES *** */
   different = 0;
   TextFace(bold);
   TextFont(font);


afkelsall:
Some context on the animation timing, since it came up while tracking this down. Not changed by this PR.

The item keep animation blocks for about 105 ms typically, and occasionally about 230 ms. The baseline is present overhead rather than drawing: every QuickDraw primitive calls `recomposite_from_window`, which composites the window stack and then presents the whole screen (`SDL_CreateTextureFromSurface` over the full logical screen, `SDL_RenderTexture`, `SDL_RenderPresent`, `SDL_SyncWindow`). The animation issues 4 `FrameOval` calls per frame over 24 frames, so one pickup costs roughly 98 full screen composite and present cycles, which works out to about 1.1 ms each.

The slow ones are not a stall. Logging the slowest single frame showed it never exceeded 15 ms, so all 24 frames are uniformly slower and the per present cost is simply doubling to around 2.4 ms for a second or so before recovering. Nothing in the composite path varies that way, so something else, maybe environmental is causing the spike). With about 98 presents per pickup, any jitter in the present pipeline gets multiplied.

Worth its own change rather than this one: presenting once per animation frame instead of once per primitive would bring the animation down to roughly 25 ms and make it much less sensitive to a slow present pipeline.

ancestral:
The treasure selection screen is painfully slow. Anything to help speed it up would be greatly appreciated.

iSynic:
> The treasure selection screen is painfully slow. Anything to help speed it up would be greatly appreciated.

Are you playing the beta 4 Release or building your own? I generally don't have an issue with slow looting except on Debug builds.

ancestral:
Beta release.

Here are 4 screen videos, Realmz 7 and 8 on the Mac, Realmz 7 on Windows, and the Beta. The Beta is the slowest! (Is the animation the culprit here?)

[treasure-recordings.zip](https://github.com/user-attachments/files/30868032/treasure-recordings.zip)

iSynic:
> Beta release.
> 
> Here are 4 screen videos, Realmz 7 and 8 on the Mac, Realmz 7 on Windows, and the Beta. The Beta is the slowest! (Is the animation the culprit here?)
> 
> [treasure-recordings.zip](https://github.c

https://github.com/user-attachments/assets/ad50132a-caf5-4ee7-a9c1-e760e9cb8dd6

Yeah, I don't know what I was thinking. It is a problem on the beta still. It does seem fixed in this branch.

PR #309, opened 2026-07-27
add marks to Mac popups

### LINKED ISSUE #202: Implement `SetItemMark` for checked menu items

chromancer:
issue: Current Native mac popup menus don't have marks, just icons. #235 added marks *only* in menu bar for Mac and #230 added icons for menus and popups. I meant to come back for popups but never did. Last nail and closes #202.

Mac only as Windows has handled this separately in #264.

fix:

- apply menu bar mapping to popup rows: checked flag or diamond mark (19) → NSControlStateValueOn (check), any other mark → NSControlStateValueMixed (dash)
- add constant kDiamondMark

effects:

- equipped items, the source character, and permanent conditions show check.
- temporary conditions show dash.
- monster items revealed by Detect Magic show check.
- rows with portraits show mark and icon together.

notes: I wonder about using diamonds from SF Symbols later. I'll test how it looks.

testing: built on Mac 26.5.2. Verified visually.

### DIFF
--- a/src/macos/MenuController.mm
+++ b/src/macos/MenuController.mm
@@ -13,6 +13,8 @@ NSMenu* MCCreateSubMenu(NSString* title, const Menu& menuRes, const std::list<st
 // We have to forward declare because of PixMap collision w/QuickDraw
 phosg::ImageRGBA8888N DecodeCIconImage(int16_t iconID);
 
+static constexpr char kDiamondMark = 19;
+
 // Wraps a decoded cicn as an NSImage, NN scale 2x, byteswap BGRA->RGBA.
 static constexpr NSInteger kMenuIconUpscale = 2;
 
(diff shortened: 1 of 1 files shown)


src/macos/MenuController.mm
```diff
@@ -13,6 +13,8 @@
 // We have to forward declare because of PixMap collision w/QuickDraw
 phosg::ImageRGBA8888N DecodeCIconImage(int16_t iconID);
 
+static constexpr char kDiamondMark = 19;
```
afkelsall, line 16: Not for this PR, but this constant is duplicated between the mac and windows versions (as `kPopupDiamondMark` in windows code). Not sure if we want to share consts between the platforms. We can think about this down the track.

afkelsall (approved):
This looks good to me. There's still things to do to get the windows experience closer to the mac experience and I appreciate all the work so far :)

PR #315, opened 2026-08-04
draw editText frame as dlog manager did

chromancer:
In discussion with @iSynic of trying to implement #301 faithfully we need the Dialog Manager's automatic editText frame per Toolbox Essentials FrameRect. This PR adds that capability, refactoring some text drawing to make sure frames only draw when appropriate.

In Classic the frame draws during item draws in DITL order. Realmz's DITLs usually put interface after text items. Those seen are because SetDialogItemText redraws item on top. Drawing it on every render puts frames on e.g., the Name modal in char gen, where original Realmz would have only shown it for a split second.

fix:

* draw frame in the EDIT_TEXT item draw
* split the TextEdit-role drawing into its own text renderer (erase + text + caret) — caret blink, focus, typing, and TEUpdateUnstyled use it, since TE never drew chrome
* use text renderer in DrawDialog

results:

* map note and spell description get their yellow outlines back
* other edit text items which draw interface after the frame don't show the frame

Built on Mac, verified correct visual behavior.

### DIFF
--- a/src/WindowManager.cpp
+++ b/src/WindowManager.cpp
@@ -542,18 +542,10 @@ public:
         break;
       }
       case ResourceFile::DecodedDialogItem::Type::EDIT_TEXT: {
-        if (!port.draw_text(text, this->rect)) {
-          wm_log.error_f("Error when rendering editable text item {}: {}", resource_id, SDL_GetError());
-        }
-
-        // Draw caret if this item is focused
-        auto window = this->owner_window.lock();
-        if (window && window->is_text_caret_visible() && window->get_focused_item().get() == this) {
-          int16_t caret_x = this->rect.left + port.measure_text(text) + 1;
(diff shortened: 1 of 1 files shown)


iSynic (approved):
This looks good to me. 

Full dialog-item rendering can draw the editText frame, but the TextEdit/caret refresh paths only redraw the text. That should let spell descriptions and map notes get the Classic frame when Realmz fills the item with `SetDialogItemText`, without making unrelated text-entry dialogs keep showing frames that Classic would have buried.

I also confirmed the map note path goes through `MyrPascalDiStr` -> `SetDialogItemText`, so this should let #301 drop the hand-drawn `FrameRect` and stay focused on the map note dialog/placement change.

Built locally on Windows and didn’t see any mechanical issues.

Issue #113, opened 2025-08-05
Screen gets fuzzy (bitmap text gets duller) sometimes when switching focus away from Realmz then back, and moving the mouse over the screen


danapplegate:
Looking a bit more carefully at this, it looks like what might be happening is that the screen is re-rendering, but the port is shifted to the right by a pixel. Then, it looks like this might perhaps be interacting poorly with the high-DPI display of my MacBook, where the shifted image falls across multiple native pixels, causing some sort of interpolation that has the effect of blurring the image.

I can't seem to reproduce it while screen recording is happening – Maybe Quicktime is disabling this pixel interpolation feature so that recordings are crisp? Not sure.

To reproduce, if you visit a tile that draws a bitmap font message, like the city gates where the guards hand you a map, the blurring should happen after about 2 seconds. Then, every time you move, the text is clear again, but then blurs again after about 2 seconds. So, it definitely seems to be triggered by the regular game loop with the delay?

jpetrie:
I tried to reproduce this in the latest commit, and I don't see any blurring on the "town patrol waves you through" message after waiting several seconds with Realmz tabbed in and out. I expect this has been addressed in one of the changes that touched rendering recently - I'd recommend we close this unless it still reproduces for you @danapplegate

danapplegate:
It's so strange, I'm still seeing this. I captured a video, you can see the lines in the pentagrams get thicker and the textures and text get fuzzier after a second or so of standing still: https://photos.app.goo.gl/CnvkyDz7fVSucH2j6

I've also noticed a couple of things:
- It only happens on my laptop's integrated retina screen (Macbook pro M3 with built-in liquid retina XDR display). It doesn't seem to happen when I connect to my monitor (Samsung C34J79x).
- If I enable persistent search, it seems that the animation of the search button somehow keeps the display refreshed and crisp.
- At some point, @fuzziqersoftware asked ChatGPT, and it came up with a plausible-sounding explanation about Metal caching an application's screen state if it hasn't received a draw call in a while. Perhaps something about the HDMI connection to the monitor prevents the display driver from doing this optimization?

ancestral:
This happens to me too. macOS Tahoe 26.5.2, MacBook Pro M3 Pro. It’s definitely some kind of anti-aliasing, I wager it’s something macOS is doing.

Current workaround: issue does not happen when you are in full screen.

I have been using Beta 4.

iSynic:
[...]

PR #316, opened 2026-08-07
Enable high pixel density rendering

iSynic:
Addresses #113.

## What changed

- Request a high-pixel-density drawable when creating the SDL window.
- Mark the macOS application bundle as high-resolution capable.

## Why

Realmz currently creates a normal-density window even on Retina displays, leaving macOS to scale the rendered output to the display's backing resolution. This appears to be why the image becomes softer or thicker after sitting idle on affected Retina displays, while continuous animation keeps it crisp.

This keeps Realmz's existing 800x600 logical presentation and apparent window size. It only allows SDL to provide a higher-resolution drawable on displays that support it; normal-density displays should continue using a 1x drawable.

## Verification

- Windows build passes.
- Built and validated a universal macOS Release test DMG containing this change.
- Tested by Discord user `mrproud`, who reports that this solution works.

### DIFF
--- a/bundle/Info.plist.in
+++ b/bundle/Info.plist.in
@@ -28,5 +28,7 @@
     <string>@PROJECT_COPYRIGHT@</string>
     <key>LSMinimumSystemVersion</key>
     <string>13.3</string>
+    <key>NSHighResolutionCapable</key>
+    <true/>
 </dict>
 </plist>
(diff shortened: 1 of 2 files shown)


chromancer (approved):
Couldn't repro issue myself, but the Discord report is that this completely fixes affected Macs (which appear to apply some power-saving resample to 1x layers when idle?).

Builds clean. drawable now reports 1600x1200, logical size and coord conversion unaffected, other platforms 1x anyway. Approved.

PR #318, opened 2026-08-07
use left and right mod keys

### LINKED ISSUE #132: Option-clicking an item in inventory is broken

chromancer:
issue: Closes #132. option, shift, and cmd are using only SDL LALT, LCTRL, LSHIFT, etc.

fix: use SDL's combined masks (SDL_KMOD_ALT/SHIFT/CTRL) for generic

effects: holding left or right mod keys works for all purposes

testing: built on Mac, verified right and left mod keys now work.

### DIFF
--- a/src/EventManager.cpp
+++ b/src/EventManager.cpp
@@ -456,11 +456,11 @@ protected:
         // control-character KCHR table and break Command-key shortcuts and typed characters.
         this->set_modifier_value(EVMOD_CONTROL_KEY_DOWN, false);
 #else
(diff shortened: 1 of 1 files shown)


iSynic (approved):
Looks good. Builds cleanly on Windows.

I think the only thing worth mentioning is that on Windows, Control is mapped to the Classic Command key separately. That mapping still checks SDL_KMOD_LCTRL, so right Control still is not equivalent there. That could be addressed in a follow-up PR if there's interest.

chromancer:
per discord discussion, windows users should get both keys too.

